PR #3, opened 2016-08-03
Add rspec, remove test-unit, erb2haml, 

joshuatobin:
* Adding rspec
* Removing test-unit
* Migrating erb to haml
* Adding factorygirl

CC: @heroku/sre


spec/spec_helper.rb
```diff
@@ -0,0 +1,42 @@
...
+Dir[Rails.root.join("spec/support/**/*.rb")].each {|f| require f}
+Dir[Rails.root.join("spec/config/**/*.rb")].each  {|f| require f}
+
+RSpec.configure do |config|
+  config.expect_with :rspec do |expectations|
+    expectations.include_chain_clauses_in_custom_matcher_descriptions = true
+  end
+
+  config.expect_with :rspec do |c|
+    c.syntax = [:should, :expect]
```
reidmix: just use expect since we don't have to support legacy syntax \o/

spec/spec_helper.rb
```diff
@@ -0,0 +1,42 @@
...
+  end
+
+  config.mock_with :rspec do |c|
+    c.syntax = :should
```
reidmix: Same

reidmix:
Awesome.

reidmix:
I **think** expect is the default/preferred method, anyway, just use the latest and make sure its true for mocks too.

joshuatobin:
@reidmix All fixed.

reidmix:
Woo 👍

PR #4, opened 2016-08-03
Stub out incidents sync endpoint

joshuatobin:
Stubbing out the sync endpoint to accept a webhook to start a sync. Depends on https://github.com/heroku/retrodot/pull/3 for tests.


joshuatobin:
@reidmix @lexelby This should be all sorted for now.

lexelby:
@joshuatobin Sorry, I wasn't clear.  Status will post _only_ the update hash, not the entire incident hash.  So you would see something like this posted:

```
     {
        "id": 3227,
        "created_at": "2016-08-03T21:24:21.541Z",
        "updated_at": "2016-08-03T21:24:23.822Z",
        "incident_id": 928,
        "contents": "We have received reports of increased failure rates during Postgres provisions. We are investigating the issue.",
        "title": "Issues with Heroku Postgres provisioning",
        "update_type": "issue",
        "services": [
          "Heroku Postgres"
        ]
      }
```

joshuatobin:
@lexelby Should be all set.

config/routes.rb
```diff
@@ -1,2 +1,5 @@
 Rails.application.routes.draw do
+  post "/incidents/sync" => "incidents#sync"
```
reidmix: Try it with the `resources` format.  There's a way to specify a single (REST) route only, and a way to define a non-REST route too.

app/controllers/incidents_controller.rb
```diff
@@ -0,0 +1,14 @@
+class IncidentsController < ApplicationController
+  protect_from_forgery except: :sync
+
+  def index
+    render text: "Welcome to Retrodot"
+  end
+
+  def sync
```
reidmix: I can help on this too, I saw the webhook PR: heroku/status#740

lexelby:
:+1:

PR #5, opened 2016-08-05
Add incident/retrospective/remediation models and migrations

joshuatobin:
This is mostly exploratory for what models we should stub out. 

CC: @heroku/sre


app/models/incident.rb
```diff
@@ -0,0 +1,6 @@
+class Incident < ActiveRecord::Base
+  has_many :retrospectives
+  has_many :remediations, through: :retrospectives
+
+  validates :title, presence: true
```
lexelby: This is going to sound a bit weird, but I think we shouldn't validate the presence of the title.  Here's the scenario I'm thinking of:

1. Something fishy starts going on, an engineer pages IC.
2. IC does `h start incident`.
3. Retrodot sees that no incident is open right now so assumes the IC is declaring the start of a new incident that hasn't been posted to the status site yet.
4. Retrodot starts a new timeline for a new incident.
5. IC has support post the incident to the status site.

In other words, it may be necessary to create an incident object without a title.

db/migrate/20160805132331_create_incidents.rb
```diff
@@ -0,0 +1,30 @@
+class CreateIncidents < ActiveRecord::Migration[5.0]
+  def change
+    create_table :incidents do |t|
+      t.date       :date
+      t.string     :title
+      t.integer    :issue
```
lexelby: Can we set this as the primary key?

app/models/incident.rb
joshuatobin: I guess the sync would just then eventually fill in the title?

db/migrate/20160805132331_create_incidents.rb
```diff
@@ -0,0 +1,30 @@
...
+
+      t.timestamps null: false
+    end
+
+    create_table :remediations do |t|
+      t.date       :date
+      t.string     :description
```
lexelby: can we add a text-search index on this?

db/migrate/20160805132331_create_incidents.rb
```diff
@@ -0,0 +1,30 @@
...
+      t.boolean    :completed
+      t.string     :pull_request
+      t.timestamps null: false
+
+      t.belongs_to :retrospective
+    end
+
+    create_table :retrospectives do |t|
+      t.date       :date
+      t.string     :description
```
lexelby: text-search here too

app/models/incident.rb
lexelby: Yup.

lexelby:
Let's go ahead and add fields that are currently in [Past Incidents](https://docs.google.com/spreadsheets/d/1d4wYxLD42ickxauV9NJEdPmAKkijIk9Gc1JjT5kYKdM/edit#gid=0):

* date of retro
* category (probably join to a categories table)
* primary team (join to a teams table, fill in with the teams we have now, or use DW data if we can)
* needs public followup
* date of public followup
* primary component (use DW data, or we can just leave this blank for now, maybe?)

At the very least, we should have date of retro, category, and public followup stuff to achieve parity with Past Incidents.

app/models/incident.rb
joshuatobin: Fixed in 3278bc1

db/migrate/20160805132331_create_incidents.rb
joshuatobin: @reidmix Do you have thoughts on this. Everything I'm reading suggests /not/ doing this. Is it a common rails practice?

db/migrate/20160805132331_create_incidents.rb
joshuatobin: @lexelby I've not done this before. Got an example?

db/migrate/20160805132331_create_incidents.rb
jmervine: Yes and no. I would remove this field all together and use the default primary key of `id` as the issue id.

joshuatobin: @jmervine Either way we need that field to record the issue # from status.heroku.com

jmervine: Ah, I see. Then yeah, it shouldn't be the primary key, although indexed for sure.

app/models/retrospective.rb
```diff
@@ -0,0 +1,6 @@
+class Retrospective < ActiveRecord::Base
+  has_many   :remediations
+  belongs_to :incident
+
+  validates :incident_id, :date, :description, presence: true
```
reidmix: You can validate on `:incident` which usually makes things easier down the line.

db/migrate/20160805132331_create_incidents.rb
reidmix: So, typically (not always), I let the table have it's own (artificial) PK.  Then I reference the external number (I might even call it `:status_id` or `:status_incident_id` to be super explicit).  

The reason I wouldn't use the external ID it as the PK is it paints you into a corner, esp. when rails gets super finicky about IDs, but also because then you get very beholden to another external system's way of numbering things as your way of numbering things.

db/migrate/20160805132331_create_incidents.rb
```diff
@@ -0,0 +1,30 @@
+class CreateIncidents < ActiveRecord::Migration[5.0]
+  def change
+    create_table :incidents do |t|
+      t.date       :date
```
reidmix: Is this the starting date of the incident?  Can naming better reflect it?

db/migrate/20160805132331_create_incidents.rb
```diff
@@ -0,0 +1,30 @@
+class CreateIncidents < ActiveRecord::Migration[5.0]
```
reidmix: I might be inclined to break these out in separate migrations -- but that's me -- maybe have the title better reflect the migration?

db/migrate/20160805132331_create_incidents.rb
```diff
@@ -0,0 +1,30 @@
...
+  def change
+    create_table :incidents do |t|
```
reidmix: I think you need `duration` as well -- are there other fields?

db/migrate/20160805132331_create_incidents.rb
```diff
@@ -0,0 +1,30 @@
...
+      t.date       :date
+      t.string     :title
+      t.integer    :issue
+
+      t.timestamps null: false
+    end
+
+    create_table :remediations do |t|
```
reidmix: Do you need remediations at this time?
I'd say boot it to keep things simple for now.

db/migrate/20160805132331_create_incidents.rb
reidmix: @lexelby what's text-search? please to teach :D

db/migrate/20160805132331_create_incidents.rb
```diff
@@ -0,0 +1,30 @@
...
+      t.belongs_to :retrospective
+    end
+
+    create_table :retrospectives do |t|
+      t.date       :date
+      t.string     :description
+      t.timestamps null: false
+
+      t.belongs_to :incident
+      t.belongs_to :remediation
```
reidmix: Nice, I always forget about these, very nice.

reidmix:
Added my feedback, great start!

db/migrate/20160805132331_create_incidents.rb
joshuatobin: @reidmix 👍 Done. I created 4 separate migrations.

db/migrate/20160805132331_create_incidents.rb
joshuatobin: Fixed as well.

joshuatobin:
@lexelby I've added a categories table and updated incidents to reflect the missing columns. I'm going to deal with teams and components separately in another PR. I need to get this PR in as its blocking on the syncher.

db/migrate/20160805132331_create_incidents.rb
```diff
@@ -0,0 +1,15 @@
+class CreateIncidents < ActiveRecord::Migration[5.0]
+  def change
+    create_table :incidents do |t|
+      t.date       :created
+      t.date       :resolved
```
reidmix: Are these a date or a date/time (or does date include time here -- i forget).

reidmix: You can use the rails naming conventions here too, but probably need to rename `:created` because of timestamps naming: `:started_on`, `:resolved_on`

db/migrate/20160805132331_create_incidents.rb
```diff
@@ -0,0 +1,15 @@
...
+      t.integer    :duration
+      t.string     :state
+      t.string     :title
+      t.integer    :issue
```
reidmix: I'd probably call this `:status_id` or `:status_incident_id` (but at least add "_id").

db/migrate/20160805132331_create_incidents.rb
```diff
@@ -0,0 +1,15 @@
...
+      t.boolean    :public_followup
+      t.date       :public_followup_date
```
reidmix: Here's a good example of rails naming conventions with a date, I'd say `:public_followup_on`

reidmix: Personally, when naming, I don't usually like putting the type in the name of the variable/column_name/etc. 
Mostly because its duplicative, but I don't think it adds anything necessarily.

db/migrate/20160809132715_create_remediations.rb
```diff
@@ -0,0 +1,13 @@
+class CreateRemediations < ActiveRecord::Migration[5.0]
+  def change
+    create_table :remediations do |t|
+      t.date       :date
```
reidmix: What is it a date of?  Is this the created date, the due date, etc.? 😄

db/migrate/20160809132715_create_remediations.rb
```diff
@@ -0,0 +1,13 @@
...
+      t.string     :description
+      t.boolean    :completed
+      t.string     :pull_request
```
reidmix: Oh interesting.  I like this.

db/migrate/20160809132722_create_retrospectives.rb
```diff
@@ -0,0 +1,12 @@
+class CreateRetrospectives < ActiveRecord::Migration[5.0]
+  def change
+    create_table :retrospectives do |t|
+      t.date       :date
+      t.string     :description
+      t.timestamps null: false
+
+      t.belongs_to :incident
+      t.belongs_to :remediation
```
reidmix: Don't need the `remediation` association -- take a look at the `schema.rb`

reidmix: The rest of the associations look good to me.

app/models/incident.rb
```diff
@@ -0,0 +1,5 @@
+class Incident < ActiveRecord::Base
+  has_many :retrospectives
+  has_many :remediations, through: :retrospectives
+  has_one  :categories
```
reidmix: I think you're looking for:
```ruby
has_one :category
```

Rails should do the right thing.

app/models/category.rb
```diff
@@ -0,0 +1,4 @@
+class Category < ActiveRecord::Base
+  has_many :incidents
+  belongs_to :incidents
+end
```
reidmix: I believe you just need:
```ruby
belongs_to :incident
```
Also, singular

"`belongs_to` will place the foreign key in the declaring model whereas `has_one` will place it in the other model."

app/models/incident.rb
reidmix: If the usecase is that an incident can only have a single category I think you want:
```ruby
belongs_to :category
```
It's also singular, rails will do the right thing.

app/models/category.rb
```diff
@@ -0,0 +1,4 @@
+class Category < ActiveRecord::Base
+  has_many :incidents
+  belongs_to :incidents
```
reidmix: If the usecase is that an incident can only have a single category, you'll just need:
```ruby
has_many :incident
```

db/migrate/20160809134428_create_categories.rb
```diff
@@ -0,0 +1,9 @@
+class CreateCategories < ActiveRecord::Migration[5.0]
+  def change
+    create_table :categories do |t|
+      t.string   :name
+
+      t.belongs_to :incident
```
reidmix: If the usecase is that an incident can only have a single category, this'll need to be a `has_many`
which then matches the model association too (I know you were working through some ambiguity there)

db/migrate/20160809132722_create_retrospectives.rb
reidmix: You probably need to add (move) the category association here:
```ruby
t.belongs_to :category
```
And remove it from the "categories" migration

db/migrate/20160809134428_create_categories.rb
reidmix: Remove (move) this, and put it in the "retrospectives" migration.

db/migrate/20160809132722_create_retrospectives.rb
reidmix: You want to do this because you want the FK to be in this table, so that each Incident has only one Category, and that any Category can be set on an Incident.

reidmix:
I noticed a few more things this round, which was mostly around naming and associations.

Nice to see these tables getting fleshed out, it's exciting to see.

app/models/incident.rb
joshuatobin: Ugh I always get these messed up. Fixed in 72277ae

db/migrate/20160805132331_create_incidents.rb
joshuatobin: Noted. 👍

db/migrate/20160809132715_create_remediations.rb
joshuatobin: Fixed in 81269dd

db/migrate/20160805132331_create_incidents.rb
joshuatobin: Fixed in 81269dd

db/migrate/20160805132331_create_incidents.rb
joshuatobin: Fixed in 81269dd

db/migrate/20160805132331_create_incidents.rb
joshuatobin: Fixed in 81269dd

db/migrate/20160809132722_create_retrospectives.rb
joshuatobin: @reidmix Would I need to remove the `belongs_to :category` in the incident model as well?

joshuatobin: Fixed in bff803d

joshuatobin:
@reidmix I think I've addressed all your suggestions

app/models/remediations.rb
```diff
@@ -0,0 +1,5 @@
+class Remediation < ActiveRecord::Base
+  belongs_to :incident
```
lexelby: This is probably my lack of activerecord knowledge talking, but why do we need the `belongs_to :incident` when we have `has_one :incident, through: :retrospective`?

db/migrate/20160805132331_create_incidents.rb
```diff
@@ -0,0 +1,15 @@
+class CreateIncidents < ActiveRecord::Migration[5.0]
+  def change
+    create_table :incidents do |t|
+      t.datetime   :started_at
+      t.datetime   :resolved_at
+      t.integer    :duration
+      t.string     :state
+      t.string     :title
+      t.integer    :status_incident_id
```
lexelby: Hmm, are you thinking that this won't be the same as the primary key ID in retrodot?  I'm concerned that having two separate IDs for an incident may lead to confusion and coding errors.  IMO having `incident.id` in Retrodot match `incident.id` in Status is a good idea.

db/migrate/20160809132715_create_remediations.rb
```diff
@@ -0,0 +1,13 @@
+class CreateRemediations < ActiveRecord::Migration[5.0]
+  def change
+    create_table :remediations do |t|
+      t.date       :created_on
+      t.string     :description
+      t.boolean    :completed
+      t.string     :pull_request
```
lexelby: I'm not sure about this.  I would think big remediations could easily require multiple pull requests?

joshuatobin: We can change this later if thats the case.

db/migrate/20160805132331_create_incidents.rb
joshuatobin: See previous comments about this.

app/models/remediations.rb
joshuatobin: @lex yeah you're actually right. It ended up causing me grief in the console so I removed it. 


```[38] pry(main)> Remediation.create!(description: 'hihi', completed: true, retrospective_id: 2, created_on: DateTime.now)
   (0.1ms)  BEGIN
  Incident Load (0.4ms)  SELECT  "incidents".* FROM "incidents" INNER JOIN "retrospectives" ON "incidents"."id" = "retrospectives"."incident_id" WHERE "retrospectives"."id" = $1 LIMIT $
2  [["id", 2], ["LIMIT", 1]]
  Retrospective Load (0.2ms)  SELECT  "retrospectives".* FROM "retrospectives" WHERE "retrospectives"."id" = $1 LIMIT $2  [["id", 2], ["LIMIT", 1]]
   (0.1ms)  ROLLBACK
  Incident Load (0.3ms)  SELECT "incidents".* FROM "incidents" INNER JOIN "retrospectives" ON "incidents"."id" = "retrospectives"."incident_id" WHERE "retrospectives"."id" = $1  [["id",
 2]]
NoMethodError: undefined method `updated?' for #<ActiveRecord::Associations::HasOneThroughAssociation:0x007f92dc82bab0>
```

db/migrate/20160805132331_create_incidents.rb
lexelby: No idea!  I was hoping you folks knew. ;)  I think postgres might have a feature for doing full-text search, I know MySQL does.  If so, we want to use it here.  If not, we're looking at something like ElasticSearch (yikes).

joshuatobin:
```
Incident.create!(state: 'resolved', title: 'api is busted', public_followup: false, started_at: '10/12/2015', status_incident_id: '992', duration: 8888)
#<Incident:0x007f92d44d1098
 id: 1,
 started_at: Thu, 10 Dec 2015 00:00:00 UTC +00:00,
 resolved_at: nil,
 duration: 8888,
 state: "resolved",
 title: "api is busted",
 status_incident_id: 992,
 public_followup: false,
 public_followup_on: nil,
 created_at: Tue, 09 Aug 2016 19:14:14 UTC +00:00,
 updated_at: Tue, 09 Aug 2016 19:14:14 UTC +00:00>

# Create Category
Category.create!(name: 'my category')
#<Category:0x007f92da8a15f0 id: 1, name: "my category">

#Create Retro
Retrospective.create!(created_on: DateTime.now, description: 'hihi', incident_id: 1, category_id: 1)
#<Retrospective:0x007f92d4e0e868
 id: 1,
 created_on: Tue, 09 Aug 2016 19:16:16 UTC +00:00,
 description: "hihi",
 created_at: Tue, 09 Aug 2016 19:16:16 UTC +00:00,
 updated_at: Tue, 09 Aug 2016 19:16:16 UTC +00:00,
 incident_id: 1,
 category_id: 1>

# Create Remedation
Remediation.create!(created_on: DateTime.now, description: 'hihihi', completed: true, retrospective_id: 1)
id: 1,
 created_on: Tue, 09 Aug 2016 19:16:43 UTC +00:00,
 description: "hihihi",
 completed: true,
 pull_request: nil,
 created_at: Tue, 09 Aug 2016 19:16:43 UTC +00:00,
 updated_at: Tue, 09 Aug 2016 19:16:43 UTC +00:00,
 retrospective_id: 1>

# Verify 
Incident.first.retrospectives
=> [#<Retrospective:0x007f92d474c890
  id: 1,
  created_on: Tue, 09 Aug 2016 19:16:16 UTC +00:00,
  description: "hihi",
  created_at: Tue, 09 Aug 2016 19:16:16 UTC +00:00,
  updated_at: Tue, 09 Aug 2016 19:16:16 UTC +00:00,
  incident_id: 1,
  category_id: 1>]

Incident.first.retrospectives.first.incident
=> #<Incident:0x007f92d64e7238
 id: 1,
 started_at: Thu, 10 Dec 2015 00:00:00 UTC +00:00,
 resolved_at: nil,
 duration: 8888,
...

 Incident.first.remediations
[#<Remediation:0x007f92d6522dd8
  id: 1,
  created_on: Tue, 09 Aug 2016 19:16:43 UTC +00:00,
  description: "hihihi",
  completed: true,
  pull_request: nil,
  created_at: Tue, 09 Aug 2016 19:16:43 UTC +00:00,
  updated_at: Tue, 09 Aug 2016 19:16:43 UTC +00:00,
  retrospective_id: 1>]

Category.first.retrospectives
=> [#<Retrospective:0x007f92dc890690
  id: 1,
  created_on: Tue, 09 Aug 2016 19:16:16 UTC +00:00,
  description: "hihi",
  created_at: Tue, 09 Aug 2016 19:16:16 UTC +00:00,
  updated_at: Tue, 09 Aug 2016 19:16:16 UTC +00:00,
  incident_id: 1,
  category_id: 1>]
```

db/migrate/20160809132715_create_remediations.rb
lexelby: Fine with me, just wanted to call it out.

lexelby:
:+1:

app/models/remediations.rb
reidmix: Yup, 👍

db/migrate/20160805132331_create_incidents.rb
reidmix: I think string is ok.  Not sure why text-search came up?

lexelby: @reidmix I'm sure we're going to want to be able to search in incident and retrospective history.  It's something we've been lacking in the existing setup for a long time.  A standard `LIKE` query to search for single terms requires a full table scan, which may not be very fast.  I guess for the relatively small number of rows in the incidents table, though, it may be fine.

reidmix: Ah i understand now..  Yeah so when we cross that bridge there are a few routes we can take, and we can use something like elasticsearch, I see where you were going with that now.  

Agreed, strings should be fine for now, we can change things up in the future when we need to.

PR #7, opened 2016-08-10
Adding dotenv-rails to work with configerator

joshuatobin:
In order to get dotenv-rails to load properly we need to set it up in `config/application.rb` before configertor loads. Also adding a `source_url` and `.env.sample` as an example which will be used in the next PR. 

CC: @jmervine @heroku/sre


lexelby:
:+1:

config/application.rb
```diff
@@ -8,8 +7,10 @@
...
 module Retrodot
   class Application < Rails::Application
-    # Settings in config/environments/* take precedence over those specified here.
-    # Application configuration should go into files in config/initializers
-    # -- all .rb files in that directory are automatically loaded.
+
+    # We have to require dotenv before configerator is loaded
+    Dotenv::Railtie.load
+
+    require_relative 'config'
```
reidmix: So in status we load it outside the Rails::Application config cycle -- ie. we load it after boot but before 'rails/all'
That makes it available very early in the boot/configuration/initialization cycle:
https://github.com/heroku/status/blob/master/config/application.rb#L2

I **think** that will solve the DotEnv problem.

joshuatobin: @reidmix Yeah I tried it all sorts of different ways. I just tried it again to be 100% sure. Still no luck anywhere but within `Rails::Application`

reidmix: :/ Ok, I may futz with it too (and see if i can make configerator super simple to use. 
What's the usecase (like how do I repro the error?)

PR #8, opened 2016-08-11
Containing all the configerator dependencies into config

reidmix:
This lets Config to be defined before any of the Application/Environment setup. 

Also, what I also found out was that `require` configs in Status would never have worked if we had used them... i'll figure that out next. (heroku/status#745)

I think I can make Configerator dotenv aware, if its available, too (subsequent PR).

/cc @heroku/sre


jmervine:
:+1:

config/config.rb
```diff
@@ -1,3 +1,6 @@
+require 'dotenv/rails-now'
```
joshuatobin: @reidmix huh okay. I was doing a variation of this as listed here: https://github.com/bkeepers/dotenv/blob/master/lib/dotenv/rails-now.rb#L5

Thanks for fixing

reidmix: I tried to do it all through the gemfile, but I think what's happening is that I am loading Config super early (which I want) and I have access to the gems, but they haven't been loaded yet (i think).  So I have to call it here.

I think eventually, rails would have loaded `dotenv/rails-now` so you wouldn't have to had to call `Dotenv::Railtie.load` explicity but that still caused problems because it happened too late, after Config was being loaded.   Which is what you were fighting with.

PR #11, opened 2016-08-16
Add a mediator for syncing incidents from a source_url

joshuatobin:
This PR give the ability to define a `source_url` along with a set of configurable environment variables so that retrodot can update its incidents state. The mediator should be generic enough that anyone could config the env vars to have retrodot conform. 

CC: @heroku/sre


app/support/mediators/incident/syncher.rb
```diff
@@ -0,0 +1,45 @@
...
+      method:       :get,
+      expects:      200,
+      idempotent:   true,
+      )
+    end
+
+    def update_incident
+      Rails.logger.info(fn: "update_incident")
+
+      @incident = ::Incident.find_or_initialize_by(incident_id: incident_details[Config.send(:incident_id)])
```
lexelby: why not just `Config.incident_id`?

config/config.rb
```diff
@@ -10,4 +10,16 @@ module Config
 
   override :frontend_host, "retro.dev", string
   required :source_url, string
+
+  # Overrides for incident syncher
+  override :duration, "duration", string
```
lexelby: Should we try to namespace this?  Like, `:source_incident_id_field` or something?

app/support/mediators/incident/syncher.rb
```diff
@@ -0,0 +1,45 @@
...
+      details.each {|i| @incident.update_attribute(i, incident_details[Config.send(i.to_sym)]) }
```
lexelby: I love how this is so flexible!  I wonder if we can go one step further, and use [jsonpath](https://github.com/joshbuddy/jsonpath)?

config/config.rb
joshuatobin: Maybe? I'm not really sure what else to call it. I'd be inclined to deal with it if it becomes an issue.

app/support/mediators/incident/syncher.rb
joshuatobin: I had all sorts of issues with this due to string interpolation. @jmervine helped me and suggested I use `send`.

app/support/mediators/incident/syncher.rb
joshuatobin: Possibly. If we think its worth spending the effort on but maybe during the second iteration.

.env.test.sample
```diff
@@ -0,0 +1 @@
+export SOURCE_URL="https://status.localhost.com/v4/api/incidents"
```
reidmix: newline plz :D

app/support/mediators/incident/syncher.rb
```diff
@@ -0,0 +1,45 @@
...
+      Rails.logger.info(fn: "call")
+      update_incident
+
+    rescue Excon::Error
+      Rails.logging.error($!, fn: "call", at: "update_incident")
+      raise $!
+    end
+
+    def incident_details
+      @@response ||= response
```
reidmix: This is a class variable, so if there's ever a time when 2 instances run, they're both going to access this.
`@response` will work for you.

app/support/mediators/incident/syncher.rb
lexelby: Sure.

lexelby:
:+1:

app/support/mediators/incident/syncher.rb
reidmix: Wonder why rails is insisting on `::` for `Incident` -- that's annoying.

mandagill: @reidmix I remember reading why that would happen in _The Well Grounded Rubyist_ but cannot remember off the top of my head and my copy is on my desk. I can look it up later if it's really annoying you.

joshuatobin: @reidmix I tried without it. The weird part was I did have access to other constants but not Incident :/

app/support/mediators/incident/syncher.rb
joshuatobin: oh shoot.

joshuatobin: @reidmix fixed

.env.test.sample
joshuatobin: Fixed

reidmix: `incident_id="id"` needed too

app/support/mediators/incident/syncher.rb
reidmix: Don't need the `send`, `incident_details[Config.incident_id]` will work to access the field.

.env.test.sample
joshuatobin: Yeah I'm not gonna put that in there. Thought about it though.

app/support/mediators/incident/syncher.rb
```diff
@@ -0,0 +1,45 @@
...
+    end
+
+    def response
```
reidmix: Some of these methods like `response` can be private

app/support/mediators/incident/syncher.rb
```diff
@@ -0,0 +1,45 @@
...
+      expects:      200,
+      idempotent:   true,
+      )
+    end
+
+    def update_incident
+      Rails.logger.info(fn: "update_incident")
+
+      @incident = ::Incident.find_or_initialize_by(incident_id: incident_details[Config.incident_id])
+      details.each {|i| @incident.update_attribute(i, incident_details[Config.send(i.to_sym)]) }
```
reidmix: So this will cause details.count updates to the database... one suggestion is build a hash fully with all key, values then pass that to `update`.

reidmix: In fact, the new way to call it is `@incident.update(params)` (over `update_attributes`)

reidmix: You know what else might be nice is having all your mappings in one (with_indifferent_access) hash so that you can do things like: `Config.incident_mappings[i]` -- and then you don't have to much with `send` and `to_sym`.

reidmix: Also, cool!

app/support/mediators/incident/syncher.rb
```diff
@@ -0,0 +1,45 @@
...
+      @incident.update_attribute(:followup_on, followup_date)
+    end
+
+    private
+    def followup_date
+      Config.followup_days.business_days.after(DateTime.parse(incident_details[Config.started_at]))
```
reidmix: Sweet.

app/support/mediators/incident/syncher.rb
```diff
@@ -0,0 +1,45 @@
...
+      idempotent:   true,
+      )
+    end
+
+    def update_incident
+      Rails.logger.info(fn: "update_incident")
+
+      @incident = ::Incident.find_or_initialize_by(incident_id: incident_details[Config.incident_id])
+      details.each {|i| @incident.update_attribute(i, incident_details[Config.send(i.to_sym)]) }
+      @incident.update_attribute(:followup_on, followup_date)
```
reidmix: I'd also add this to the aforementioned hash before passing to `update`.

app/support/mediators/incident/syncher.rb
```diff
@@ -0,0 +1,45 @@
...
+    end
+
+    private
+    def followup_date
+      Config.followup_days.business_days.after(DateTime.parse(incident_details[Config.started_at]))
+    end
+
+    def details
+      %w[incident_id state title started_at resolved_at duration requires_followup]
```
reidmix: Could also be a `%i[]` to make them symbols.

app/support/mediators/incident/syncher.rb
```diff
@@ -0,0 +1,45 @@
...
+
+    def response
+      MultiJson.decode(get_incident.body)
+    end
+
+    def get_incident
+      Excon.new("#{Config.source_url}/#{@options[:incident]}").request(
+      method:       :get,
+      expects:      200,
+      idempotent:   true,
```
reidmix: If you would: tab these in -- that way I can see they are "under" `Excon.new`.

app/support/mediators/incident/syncher.rb
```diff
@@ -0,0 +1,45 @@
...
+    def incident_details
+      @response ||= response
+    end
+
+    def response
+      MultiJson.decode(get_incident.body)
+    end
+
+    def get_incident
+      Excon.new("#{Config.source_url}/#{@options[:incident]}").request(
```
reidmix: Do we need to specify any timeouts or retries -- ie. are there reasonable defaults?

app/support/mediators/incident/syncher.rb
```diff
@@ -0,0 +1,45 @@
+module Mediators::Incident
+  class Syncher < Mediators::Base
+    def call
+      Rails.logger.info(fn: "call")
+      update_incident
+
+    rescue Excon::Error
+      Rails.logging.error($!, fn: "call", at: "update_incident")
```
reidmix: It'd also be nice to see which `incident_id` caused the failure (to aid in debugging).

reidmix: I actually like how all the parsing/fetching will just throw an error and we catch it here -- keeps the code nice and clean and the errors are still logged and rethrown 👍

.env.test.sample
reidmix: Agreed -- it was "my" use, not the general one.  I had to think about it too. 👍

config/config.rb
```diff
@@ -9,5 +9,17 @@ module Config
...
+  override :incident_id, "incident_id", string
+  override :state, "state", string
+  override :title, "title", string
+  override :started_at, "started_at", string
+  override :resolved_at, "resolved_at", string
+  override :incident_id, "incident_id", string
+  override :requires_followup, "requires_followup", string
```
reidmix: These are the items i was thinking it might be easier to deal with if they were all in hash.

reidmix: Just an idea.

We could create a `hash(string)` caster in Configerator to make it dead-simple for configvars that looks like `incident_id:id,title:description` etc. and maybe even let a hash define the default values.

db/migrate/20160810124629_change_status_incident_id.rb
```diff
@@ -0,0 +1,5 @@
+class ChangeStatusIncidentId < ActiveRecord::Migration[5.0]
+  def change
+     rename_column :incidents, :status_incident_id, :incident_id
```
reidmix: yes!

reidmix:
@joshuatobin I really like how you have the mediator laid out -- it's very readable and understandable.  I left a bunch of comments and suggestions, the most important being around the `update_attributes`.

app/support/mediators/incident/syncher.rb
reidmix: Last suggestion, this line after refactor *might* be a good candidate as a helper method.

config/config.rb
joshuatobin: yeah possibly. I'm going to vote to punt for now. This works just fine.

app/support/mediators/incident/syncher.rb
```diff
@@ -0,0 +1,52 @@
...
+        expects:      200,
+        idempotent:   true,
+      )
+    end
+
+    def update_incident
+      Rails.logger.info(fn: "update_incident")
+
+      @incident = ::Incident.find_or_initialize_by(incident_id: parse_details[Config.incident_id])
+      @incident.update(parse_details)
```
joshuatobin: @reidmix fixed this here with `parse_details`. Let me know what you think. Also switched to using `update`

app/support/mediators/incident/syncher.rb
reidmix: Yeah, I could see a need for this when responses become nested or deeply-nested.
I think for now, everything is at the top level and it makes mapping fairly simple.

app/support/mediators/incident/syncher.rb
reidmix: @mandagill I'd expect that rails would start looking in the global namespace after it couldn't find something in the module namespaces -- I believe this is a Rails thing -- but not sure.  It's not so important, just eye-blight.  
I can follow up on it later.

reidmix: @tobin it's very weird, I see it's happening too.  No clue what's up but thanks for the hint.  No reason to block on it.  Thanks.

app/support/mediators/incident/syncher.rb
reidmix: Oh that's nice.

app/support/mediators/incident/syncher.rb
```diff
@@ -0,0 +1,52 @@
...
+    private
+    def followup_date
+      Config.followup_days.business_days.after(DateTime.parse(incident_details[Config.started_at]))
+    end
+
+    def parse_details
+      details.each_with_object({}) do |attr, hsh|
+        hsh[attr] = incident_details[attr]
+      end.with_indifferent_access
+    end
```
reidmix: This is so beautiful, it makes me so happy 💃 

Alternatively you can also do something like: `incident_details.slice(*details).with_indifferent_access`

app/support/mediators/incident/syncher.rb
```diff
@@ -0,0 +1,52 @@
...
+      )
+    end
+
+    def update_incident
+      Rails.logger.info(fn: "update_incident")
+
+      @incident = ::Incident.find_or_initialize_by(incident_id: parse_details[Config.incident_id])
+      @incident.update(parse_details)
+
+      @incident.update(followup_on: followup_date)
```
reidmix: you can merge `followup_on: followup_date` with `parse_details` on line 32.
The advantage is that it will only be one DB call instead of two.

reidmix:
Did another pass, looking good, comments inline.

app/support/mediators/incident/syncher.rb
joshuatobin: I love `each_with_object` so much.

.env.test.sample
lexelby: Eek, no! The `id` field is the _update_'s ID.  The incident ID is in `incident_id`.  That's how the v4 API presents an update object.

app/support/mediators/incident/syncher.rb
lexelby: Our policy is that a followup should be posted 7 days after the incident, regardless of business days.

app/support/mediators/incident/syncher.rb
joshuatobin: @reidmix fixed in 8848830

app/support/mediators/incident/syncher.rb
joshuatobin: @reidmix The default timeout is `60` seconds. The `idempotent:   true` means it will retry `3` times on error.

lexelby: Excon has reasonable defaults provided that we are on version > 0.50.1, I believe.

app/support/mediators/incident/syncher.rb
```diff
@@ -0,0 +1,52 @@
...
+
+    def parse_details
+      details.each_with_object({}) do |attr, hsh|
+        hsh[attr] = incident_details[attr]
+      end.with_indifferent_access
+    end
+
+    def details
+      %w[incident_id state title started_at resolved_at duration requires_followup]
+    end
```
joshuatobin: I just realized this is all wrong. This won't take advantage of the config vars if someone changes the defaults.

app/support/mediators/incident/syncher.rb
```diff
@@ -0,0 +1,60 @@
...
+        expects:      200,
+        idempotent:   true,
+      )
+    end
+
+    def update_incident
+      Rails.logger.info(fn: "update_incident")
+
+      @incident = ::Incident.find_or_initialize_by(incident_id: parse_details[:incident_id])
+      @incident.update(parse_details.merge!(followup_on: followup_date))
```
joshuatobin: @reidmix condensed the update into one call here.

app/support/mediators/incident/syncher.rb
```diff
@@ -0,0 +1,60 @@
...
+        end.with_indifferent_access
+    end
+
+    # Create a mapping of details to whatever custom field was passed in.
+    # {"incident_id" => "id"}
+    def mapping
+      details.each_with_object({}) do |attr, hsh|
+        hsh[attr] = Config.send(attr.to_sym)
+      end
+    end
```
joshuatobin: @reidmix Added mappings here so we're pulling the right key/value pair out of `incident_details`

app/support/mediators/incident/syncher.rb
joshuatobin: Fixed above with `mapping`

joshuatobin:
@reidmix this should address all your concerns now.

.env.test.sample
jmervine: No `export` needed.

Gemfile
```diff
@@ -22,11 +25,14 @@ group :development, :test do
   gem 'factory_girl_rails'
   gem 'rspec-rails'
   gem 'dotenv-rails'
```
jmervine: This not being installed in prod may raise an error, given that `dotenv/rails-now` is required in `config/config.rb`.

app/support/mediators/base.rb
```diff
@@ -0,0 +1,15 @@
+require 'configerator'
```
jmervine: This shouldn't be here. `configerator` is used to build `Config`, which should be loaded already.

app/support/mediators/base.rb
```diff
@@ -0,0 +1,15 @@
...
+
+module Mediators
+  class Base
+    include Configerator
```
jmervine: Same as above, this shouldn't be here.

config/config.rb
```diff
@@ -10,4 +10,16 @@ module Config
 
   override :frontend_host, "retro.dev", string
```
jmervine: Should this default to the development value? I would think that setting it in `.env.development` and/or `.env.test` and making it `required`.

db/migrate/20160815141857_change_public_followup.rb
```diff
@@ -0,0 +1,6 @@
+class ChangePublicFollowup < ActiveRecord::Migration[5.0]
+  def change
+      rename_column :incidents, :public_followup,    :requires_followup
+      rename_column :incidents, :public_followup_on, :followup_on
```
jmervine: Personally, I prefer `:followup_at` to match the `updated_at` and `created_at` pattern. But it really doesn't matter. >.<

spec/spec_helper.rb
```diff
@@ -1,14 +1,14 @@
 ENV["RAILS_ENV"] ||= 'test'
 ENV["HOSTNAME"] = 'retrodot.test'
```
jmervine: This should be moved to `.env.test` or under `Dotenv.load!` IMO

spec/support/mediators/incident/syncher_spec.rb
```diff
@@ -0,0 +1,49 @@
+require 'spec_helper'
+
+describe Mediators::Incident::Syncher do
+  let(:id) { '900' }
+
+  def encoded_data
+   MultiJson.encode(incident_details)
```
jmervine: Missing one indentation space here. =P

spec/spec_helper.rb
joshuatobin: @jmervine Yeah I have another PR to fix all the environment related things.

spec/support/mediators/incident/syncher_spec.rb
joshuatobin: Fixed in 906acfc

config/config.rb
joshuatobin: Yup I'll do that. going to submit another PR

app/support/mediators/incident/syncher.rb
reidmix: I think you can combine `parse_details` and `mapping` -- the difference being `mapping[attr]` vs `Config.send(attr)` -- which i think helped you at the time to figure this out.

db/migrate/20160815141857_change_public_followup.rb
reidmix: FWIW, the pattern is `_on` for dates and `_at` for datetimes.  I _believe_ this is a date.

Gemfile
reidmix: Right, good catch.

Just remove `require 'dotenv/rails-now'` from `config/config.rb` and you'll be golden, since Configerator is doing that for you and failing gracefully if `dotenv*` is not installed.

reidmix:
Added my comments, +1'd the mervine comments I agreed with :D

app/support/mediators/incident/syncher.rb
reidmix: Ah, thanks for checking and explaining. TIL 👍

app/support/mediators/incident/syncher.rb
```diff
@@ -0,0 +1,51 @@
...
+
+    def followup_date
+      Config.followup_days.business_days.after(DateTime.parse(incident_details[Config.started_at]))
+    end
+
+    def parse_details
+      @parse_details ||= \
+        details.each_with_object({}) do |attr, hsh|
+          hsh[attr] = incident_details[Config.send(attr.to_sym)]
+        end.with_indifferent_access
```
joshuatobin: @reidmix Condensed the mapping method down here.

app/support/mediators/base.rb
```diff
@@ -0,0 +1,12 @@
...
+
+    def self.run(options={})
+      new(options).call
+    end
+
+    def initialize(options={})
+      @options = options
+    end
+  end
+end
```
joshuatobin: @jmervine Removed all references to configerator.

Gemfile
joshuatobin: Fixed.

joshuatobin:
@reidmix @jmervine Made all noted changes. Merging 💃

app/support/mediators/incident/syncher.rb
reidmix: Looks good.  FWIW & BTW, `send` will take a symbol or a string :D

reidmix:
👍

PR #12, opened 2016-08-17
Adding rake task for sync mediator

joshuatobin:
This PR depends on https://github.com/heroku/retrodot/pull/11

CC: @heroku/sre


lib/tasks/sync_incident.rake
```diff
@@ -0,0 +1,6 @@
+namespace :sync do
+  desc "Sync incident via the Syncher Mediator"
+  task :incident, [:incident_id] => :enviroment do |t, args|
+    Mediators::Incident::Syncher.run(incident: args)
+  end
+end
```
reidmix: rake arguments are always mind-bendy.
I think you may want `args[:incident_id]`

reidmix: Does this include the task to sync all tasks?

Also remind me how are we thinking of using these tasks in scheduler -- my brain isn't remembering the usecase.

joshuatobin: No. I'm basically just stubbing this out for one incident atm.

joshuatobin: @reidmix This is all set.

lexelby:
:+1:

jmervine:
**If** we're going to continue with this being open source, this should probably live in a syncer gem along with the code to sync. That code can probably just be put directly in the rake task. Something like `retrodot-status-sync` as a private gem. At which point the sync functionality is completely abstracted from the app.

That opens us up to the possibility of creating `retrodot-statuspage.io-sync' and retrodot-status.io-sync` gems as well.

jmervine:
Something like...

```ruby
# file: lib/task/sync.task                                                                              
require 'excon'                                                                                                                                                                                                                                        
require 'uri'                                                                   
require 'json'                                                                  
                                                                                   
namespace :status do                                                               
  task :sync do   

    # not sure if you can do methods in a task, so making these lambdas for now                                                                 
    endpoint_for = ->(page) do                                                     
      "#{ENV['SOURCE_URL']}?per_page=100&page=#{page}"                             
    end                                                                            
                                                                                   
    # not sure if you can do methods in a task, so making these lambdas for now                                                                 
    next_in = ->(response) do                                                      
      response.headers['Link'].split(',').last.match(/<.+[?|&]page=(\d+).*>/)[1].to_i
    end                                                                            
                                                                                   
    @page = 1                                                                      
                                                                                   
    while true                                                                     
      response  = Excon.get(endpoint_for(@page))                                   
      ids = JSON.parse(response.body).map { |i| i["id"] }                          
                                                                                   
      ids.each do |id|                                                             
        # whatever you want to do with ids                                         
      end                                                                          
                                                                                   
      if (next_page = next_in(response)) > @page                                   
        @page = next_page                                                          
      else                                                                         
        break                                                                      
      end                                                                          
    end                                                                            
  end                                                                              
end                                                                                
```

You'd need to good around on how to include a rake task from a gem in a project, but it shouldn't be hard to find.

app/support/mediators/incident/syncher.rb
```diff
@@ -2,32 +2,47 @@ module Mediators::Incident
...
+      update_incident
     rescue Excon::Error
       Rails.logging.error($!, fn: "call", at: "update_incident", incident_id: @options[:incident])
       raise $!
     end
 
+    def update_incident
+      Rails.logger.info(fn: "update_incident")
+
+      if @options.is_a?(Array)
```
jmervine: I'm not sure on passing an Array of incidents, I realize it makes it easier b/c this way you either pass a Hash of options -- which includes `incident` -- or an Array incidents. But I think I would want the mediator to handle a single incident. It should still be simple b/c `@options.has_key?(:incident)` is a fetch, while `@options.has_key?(:id)` is a `find_or_create`.

app/support/mediators/incident/syncher.rb
```diff
@@ -2,32 +2,47 @@ module Mediators::Incident
...
+      else
+        @incident = ::Incident.find_or_initialize_by(incident_id: parse_details[:incident_id])
+        @incident.update(parse_details.merge!(followup_on: followup_date))
+      end
+    end
+
+    def parse_details(incident=nil)
+      @parse_details ||= \
+      details.each_with_object({}) do |attr, hsh|
+        hsh[attr] = incident_details[Config.send(attr.to_sym)]
```
jmervine: Now sure how this is working? I would assume that this would be where you would fetch the incident from status site, if an id was passed and then format.

app/support/mediators/incident/syncher.rb
```diff
@@ -2,32 +2,47 @@ module Mediators::Incident
...
+      end.with_indifferent_access
+    end
+
     def incident_details
       @response ||= response
```
jmervine: You might consider moving the check on `@options.has_key?(:id)` here? Something like...

```ruby
def incident_details
  @response ||= @options.has_key?(:id) ? @options : response
end
```

PR #13, opened 2016-08-18
Updating env.test.sample

joshuatobin:
CC: @heroku/sre


lexelby:
:+1:

.env.development.sample
```diff
@@ -0,0 +1,8 @@
+SOURCE_URL="https://status.localhost.com/v4/api/incidents"
```
jmervine: I would make SOURCE_URL examples more generic given that our status site is not open source...

```
SOURCE_URL=https://status.example.com/api/endpoint
```

Or something like that.

joshuatobin:
@jmervine Agreed. I've updated the `source_url` to be generic.

reidmix:
👍

PR #15, opened 2016-08-18
Add rake task for sync all incidents

joshuatobin:
I've refactored the mediator a bit to sync either one incident by passing an id, or passing an Array of Hashes. I'm holding off on tests until ya'll have a first pass at it. 

CC: @heroku/sre


app/support/mediators/incident/syncher.rb
```diff
@@ -2,30 +2,44 @@ module Mediators::Incident
...
     end
 
-    def incident_details
-      @response ||= response
+    def update_incident
+      Rails.logger.info(fn: "update_incident")
+
+      if @options.is_a?(Array)
+        @options.each do |incident|
+          @response = incident
```
lexelby: This seems a little weird to me, faking out the rest of the code to think it's made a web request.  I wonder if there's a different way to do this than just repeatedly setting `@response`?  @reidmix, any ideas?

app/support/mediators/incident/syncher.rb
```diff
@@ -2,30 +2,44 @@ module Mediators::Incident
   class Syncher < Mediators::Base
     def call
       Rails.logger.info(fn: "call")
-      update_incident
 
+      update_incident
     rescue Excon::Error
-      Rails.logging.error($!, fn: "call", at: "update_incident", incident_id: @options[:incident])
+      msg = {fn: "call", at: "update_incident"}
+      msg.merge!(incident_id: @options[:incident]) if @options[:incident]
```
lexelby: If `@options` is an array, this will throw `TypeError: no implicit conversion of Symbol into Integer`.

lexelby: I wonder if we can avoid having `@options` be of possibly differing types.  Instead of array vs hash, can you use a different hash key to signify that the caller has passed a group of incident bodies?  Like `@options[:incidents]`?

lib/tasks/sync_incident.rake
```diff
@@ -0,0 +1,8 @@
+namespace :sync do
+  desc "Sync incident via the Syncher Mediator"
+  task :incident, [:incident_id] => :environment do |t, args|
+    include Helpers::Paginator
+
+    args[:incident_id] ? Mediators::Incident::Syncher.run(incident: args[:incident_id]) : Mediators::Incident::Syncher.run(fetch)
```
lexelby: This is just my personal stylistic preference, but I think this might look better as an `if/else`.

app/support/mediators/incident/syncher.rb
joshuatobin: Yeah not sure. I'm def not excited how this bit came out so looking for suggestions.

lib/tasks/sync_incident.rake
lexelby: I wonder... rather than having the syncher deal with an array of incident bodies, might it make more sense for _this_ code to iterate over all of the incidents and call the syncher mediator one time per incident body?

app/support/mediators/incident/syncher.rb
joshuatobin: Yeah that would probably be better. I'm going to refactor some bits.

app/support/mediators/incident/syncher.rb
jmervine: That's just naming at the point, rename `response` to `payload` and it's less bothersome. I had the same thought, but you're not actually faking the response, just dictating that the payload is present and doesn't need to be fetched.

app/support/mediators/incident/syncher.rb
```diff
@@ -2,48 +2,55 @@ module Mediators::Incident
...
 
-    def incident_details
-      @response ||= response
+    def update_incident
+      Rails.logger.info(fn: "update_incident")
+
+      if @options[:incidents]
+        @options[:incidents].each do |incident|
+          @payload = incident
+          @incident = ::Incident.find_or_initialize_by(incident_id: incident['id'])
```
joshuatobin: @lexelby @jmervine - Re-jiggered this a bit passing in @options[:incidents] instead of just an Array. Also renamed response to payload 👯

app/support/mediators/incident/syncher.rb
```diff
@@ -2,48 +2,55 @@ module Mediators::Incident
...
 
-      @incident = ::Incident.find_or_initialize_by(incident_id: parse_details[:incident_id])
-      @incident.update(parse_details.merge!(followup_on: followup_date))
+    def get_incident
+      Excon.new("#{Config.source_url}/#{@options[:incident]}").request(method: :get, expects: 200, idempotent: true)
     end
 
     private
-    def response
+    def payload
```
lexelby: You're right, just renaming to `payload` makes all the difference.

app/support/mediators/incident/syncher.rb
```diff
@@ -2,48 +2,55 @@ module Mediators::Incident
...
+      Rails.logger.error($!, msg)
       raise $!
     end
 
-    def incident_details
-      @response ||= response
+    def update_incident
+      Rails.logger.info(fn: "update_incident")
+
+      if @options[:incidents]
```
reidmix: one thing you can do is create your own initializer that requries `:incidents` then you can remove this test.

reidmix: Then you can save it to attr_accessible too (instead of having to reach in `@options`)

reidmix: Example: https://github.com/heroku/laika/blob/master/lib/mediators/alarm/creator.rb#L3-L5

app/support/mediators/incident/syncher.rb
```diff
@@ -2,48 +2,55 @@ module Mediators::Incident
...
+      details.each_with_object({}) do |attr, hsh|
+        hsh[attr] = incident_details[Config.send(attr.to_sym)]
+      end.with_indifferent_access
     end
 
-    def update_incident
-      Rails.logger.info(fn: "update_incident")
+    def incident_details
+      @payload ||= payload
+    end
```
reidmix: 👍 maybe even call the method `payload`

app/support/mediators/incident/syncher.rb
```diff
@@ -2,48 +2,55 @@ module Mediators::Incident
...
+      @parse_details ||= \
+      details.each_with_object({}) do |attr, hsh|
+        hsh[attr] = incident_details[Config.send(attr.to_sym)]
+      end.with_indifferent_access
     end
 
-    def update_incident
-      Rails.logger.info(fn: "update_incident")
+    def incident_details
+      @payload ||= payload
```
reidmix: combine the contents of `payload` here.  Then maybe call this method `payload`

app/support/mediators/incident/syncher.rb
reidmix: So there's very little difference between `if` and `else` blocks besides how to get the `id` (where the format sets/overloads the payload so `parse_details` works).  This is tough to unravel. 

I don't have a clear answer, but what I'm thinking is if you extract "how" you retrieve the data, and pass only the data into the mediator to parse, calculate, and save the results.  That's going to make it much more simple.

lib/helpers/paginator.rb
```diff
@@ -0,0 +1,30 @@
+module Helpers
+  module Paginator
+
+    def endpoint_for(page)
+      "#{Config.source_url}?per_page=100&page=#{page}"
```
reidmix: FUTURE:  I don't think it matters for now what you have is working. 

Just be aware that building URLs like this ends up being very brittle.
I tend to use `URI.parse` and things like `Rack::Utils.parse_nested_query` and `Rack::Utils.build_nested_query` to help modify/build URLs.

app/support/mediators/incident/syncher.rb
joshuatobin: 👍

app/support/mediators/incident/syncher.rb
joshuatobin: Yeah I could do that. I agree its semi confusing logic and might just make more sense to have separate mediators. I can whip that up

app/support/mediators/incident/syncher.rb
joshuatobin: Ah yeah. Thats much cleaner and feels like hacky.

lib/helpers/paginator.rb
joshuatobin: 👍 Agreed. I need to either find or write a gem for the pagination bits.. It feels wrong including this here but it will work for now.

app/support/mediators/incident/syncher.rb
reidmix: I have ideas on how to unravel this -- I agree, not in love with the "fake out" although I understand how it came to be :D  I'm going to work with @joshuatobin on Monday.

app/support/mediators/incident/persister.rb
```diff
@@ -0,0 +1,42 @@
...
+
+    def initialize(incident:)
+      @incident_data = incident.with_indifferent_access
+    end
+
+    def call
+      Rails.logger.info(fn: "call")
+
+      update_incident
+    rescue Excon::Error
```
lexelby: I think now this exception handler belongs in one or both of `OneSyncher` and `MultiSyncher`?

lexelby:
I love the makeover! It's much cleaner now.

app/support/mediators/incident/persister.rb
joshuatobin: @lexelby good catch! Fixed in 89881ed

app/support/mediators/incident/multi_syncher.rb
```diff
@@ -6,6 +6,9 @@ def call
       fetch(Config.source_url).each do |incident|
         Mediators::Incident::Persister.run(incident: incident)
       end
+    rescue Excon::Error
+      Rails.logger.error($!, fn: "call", at: "run", incident_id: incident[Config.incident_id])
```
lexelby: I don't think `Rails.logger.error($!, ...` works.  I had to correct that in my PR because ruby complains that the `error` method accepts only one arg, and you're passing two (`$!` and the hash arg).

lexelby:
:+1:

app/support/mediators/incident/multi_syncher.rb
```diff
@@ -0,0 +1,14 @@
...
+  class MultiSyncher < Mediators::Base
+    include Helpers::Paginator
+
+    def call
+      fetch(Config.source_url).each do |incident|
+        Mediators::Incident::Persister.run(incident: incident)
+      end
+    rescue Excon::Error
+      Rails.logger.error($!, fn: "call", at: "run", incident_id: incident[Config.incident_id])
+      raise $!
```
reidmix: `raise` here without an argument will re-raise what comes in on the `rescue`

app/support/mediators/incident/one_syncher.rb
```diff
@@ -0,0 +1,26 @@
...
+
+    def initialize(id:)
+      @id = id
+    end
+
+    def call
+      Mediators::Incident::Persister.run(incident: payload)
+    rescue Excon::Error
+      Rails.logger.error(error: $!, fn: "call", at: "run", incident_id: id)
+      raise $!
```
reidmix: Ditto.

app/support/mediators/incident/one_syncher.rb
```diff
@@ -0,0 +1,26 @@
...
+    end
+
+    private
+    def payload
```
reidmix: small tiny thing: you could probably call this `incident` here (matches MultiSyncer)

app/support/mediators/incident/persister.rb
```diff
@@ -0,0 +1,36 @@
...
+    def update_incident
+      Rails.logger.info(fn: "update_incident")
+      incident = ::Incident.find_or_initialize_by(incident_id: parse_details[:incident_id])
+      incident.update(parse_details.merge!(followup_on: followup_date))
+    end
+
+    def parse_details
+      @parse_details ||= \
+      details.each_with_object({}) do |attr, hsh|
+        hsh[attr] = incident_data[Config.send(attr.to_sym)]
```
reidmix: FYI, don't need the `to_sym`

app/support/mediators/incident/persister.rb
```diff
@@ -0,0 +1,36 @@
...
+      end.with_indifferent_access
+    end
+
+    private
+    def followup_date
+      Config.followup_days.business_days.after(DateTime.parse(parse_details[:started_at]))
```
reidmix: What did we decide about business days?  It could be a config whether to use business days or not.

spec/support/mediators/incident/multi_syncher_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+      Mediators::Incident::MultiSyncher.run
+
+      incident1.reload
+      incident2.reload
+
+      assert_equal "resolved", incident1.state
+      assert_equal "resolved", incident2.state
+    end
+  end
+end
```
reidmix: I'd say there should be tests that iterates over pages too. (to test pagination -- or tests for the paginator is probably more accurate).

reidmix:
Added a few minor comments.  I'd say the one that's most important is tests needed for the paginator.

app/support/mediators/incident/persister.rb
joshuatobin: I don't remember anyone bringing it up. I guess it could be a config :/

reidmix: OMG maybe I'm imagining it :D @lexelby did you say that we usually have to have a followup in consecutive days?

lexelby: Yup, I mentioned it on slack.

PR #16, opened 2016-08-18
Add endpoint for to sync incidents via a webhook.

joshuatobin:
CC: @heroku/sre


app/controllers/incidents_controller.rb
```diff
@@ -6,9 +6,17 @@ def index
...
   def sync
-    logger.info(ns: self.class.name, fn: :sync, incident: params["incident_id"])
+    logger.info(ns: self.class.name, fn: :sync, incident: check_params["incident_id"])
 
-    #TODO Call the sync worker
-    head :ok
+    if Mediators::Incident::Syncher.run(incident: params['incident_id'])
+      render json: nil, status: 202
+    else
+      render json: nil, status: 500
```
lexelby: Interestingly, 500 will cause Status to retry the webhook.

spec/controllers/incidents_controller_spec.rb
```diff
@@ -1,26 +1,34 @@
...
-             }
+    {
+      "incident_id": 3227,
+      "started_at": "2016-08-03T21:24:21.541Z",
+      "resolved_at": "2016-08-03T21:24:23.822Z",
+      "contents": "We have received reports of increased failure rates during Postgres provisions. We are investigating the issue.",
+      "title": "Issues with Heroku Postgres provisioning",
+      "duration": 6000,
+      "state": "open",
+      "requires_followup": true,
```
lexelby: I don't think this goes here.  This JSON object is still just the serialized representation of an Update, and I don't think individual Updates have `requires_followup`, right?

joshuatobin: Yeah it doesn't really matter. If you look at the controller you can pass it a payload of anything, as long as it has 'incident_id'. Let me write another tests that also fails

app/controllers/incidents_controller.rb
joshuatobin: hrm is that a good thing? I would think so?

lexelby: Yup, I was just commenting on that fact in case you weren't aware.  It's a nice little added bonus for free! :)

spec/controllers/incidents_controller_spec.rb
lexelby: Ah, fair point.

joshuatobin: @lexelby Added another spec for invalid data.

lexelby:
:+1:

PR #22, opened 2016-08-23
Followup -> Review

joshuatobin:
CC: @heroku/sre


lexelby:
:+1:

lexelby:
Should we add `maintain_test_schema!` to status and other projects?

joshuatobin:
@lexelby Yeah I found `maintain_test_schema!` to save me a bit of hassle.

.env.test.sample
```diff
@@ -4,5 +4,5 @@ STARTED_AT="started_at"
 INCIDENT_ID="incident_id"
 STATE="state"
 RESOLVED_AT="resolved_at"
-REQUIRES_FOLLOWUP="requires_followup"
+review="review"
```
reidmix: s/review/REVIEW

spec/spec_helper.rb
```diff
@@ -10,6 +10,9 @@
 require 'webmock/rspec'
 require 'database_cleaner'
 
+# Checks for pending migrations before tests are run.
+ActiveRecord::Migration.maintain_test_schema! if defined?(ActiveRecord::Migration)
```
reidmix: Weird this isn't the default?

lexelby: Clearly, the default is not to ensure that the test db's schema is migrated so that you can test whether your code is backward-compatible with the old schema. :troll:

In other words, I totally agree with you.

joshuatobin: Yeah I'm not sure,but fwiw this just save me once again running specs in development.

JuanitoFatas: I guess the RSpec setup not ran by `$ rails generate rspec:install`, normally it will create a `spec/spec_helper.rb` and `spec/rails_helper.rb`. And this line should be generated in `spec/rails_helper.rb`.

Generated `spec/rails_helper.rb` example:

```ruby
# This file is copied to spec/ when you run 'rails generate rspec:install'
ENV['RAILS_ENV'] ||= 'test'
require File.expand_path('../../config/environment', __FILE__)
# Prevent database truncation if the environment is production
abort("The Rails environment is running in production mode!") if Rails.env.production?
require 'spec_helper'
require 'rspec/rails'
# Add additional requires below this line. Rails is not loaded until this point!

# Requires supporting ruby files with custom matchers and macros, etc, in
# spec/support/ and its subdirectories. Files matching `spec/**/*_spec.rb` are
# run as spec files by default. This means that files in spec/support that end
# in _spec.rb will both be required and run as specs, causing the specs to be
# run twice. It is recommended that you do not name files matching this glob to
# end with _spec.rb. You can configure this pattern with the --pattern
# option on the command line or in ~/.rspec, .rspec or `.rspec-local`.
#
# The following line is provided for convenience purposes. It has the downside
# of increasing the boot-up time by auto-requiring all files in the support
# directory. Alternatively, in the individual `*_spec.rb` files, manually
# require only the support files necessary.
#
# Dir[Rails.root.join('spec/support/**/*.rb')].each { |f| require f }

# Checks for pending migration and applies them before tests are run.
# If you are not using ActiveRecord, you can remove this line.
ActiveRecord::Migration.maintain_test_schema!

RSpec.configure do |config|
  # Remove this line if you're not using ActiveRecord or ActiveRecord fixtures
  config.fixture_path = "#{::Rails.root}/spec/fixtures"

  # If you're not using ActiveRecord, or you'd prefer not to run each of your
  # examples within a transaction, remove the following line or assign false
  # instead of true.
  config.use_transactional_fixtures = true

  # RSpec Rails can automatically mix in different behaviours to your tests
  # based on their file location, for example enabling you to call `get` and
  # `post` in specs under `spec/controllers`.
  #
  # You can disable this behaviour by removing the line below, and instead
  # explicitly tag your specs with their type, e.g.:
  #
  #     RSpec.describe UsersController, :type => :controller do
  #       # ...
  #     end
  #
  # The different available types are documented in the features, such as in
  # https://relishapp.com/rspec/rspec-rails/docs
  config.infer_spec_type_from_file_location!

  # Filter lines from Rails gems in backtraces.
  config.filter_rails_from_backtrace!
  # arbitrary gems may also be filtered via:
  # config.filter_gems_from_backtrace("gem name")
end
```

reidmix: We should remove the ` if defined?(ActiveRecord::Migration)` bit, I'm sure we'll always have Migrations :D

reidmix: Looks like the only thing missing from the generated rspec is `config.infer_spec_type_from_file_location!` which is very handy, and I'd add that back in.

reidmix: FWIW, I'm not in love with `rails_helper` and `require spec_helper` -- I mean the rspec folks always seem to make things unnecc. complicated :(

JuanitoFatas: > I'm not in love with rails_helper and require spec_helper -- I mean the rspec folks always seem to make things unnecc

I think they separated `spec_helper.rb` and `rails_helper.rb` is that sometimes we have tests that do not need to load rails. So they put RSpec specific (& what every spec needs to load) in `spec_helper.rb` and Rails + Rails-specific configs in `rails_helper.rb`. :)

reidmix: That makes sense, but still. 😄

jmervine: I like that they provide both out of the box, but I'm also +1 for merging in use-cases like this one. =P

PR #25, opened 2016-08-24
Only add followup dates if review is true

joshuatobin:
CC: @heroku/sre


lexelby:
:+1:

reidmix:
Looks good... you _could_ have the business days as a config -- but i think it might be YAGNI at this point too.

jmervine:
👍

PR #26, opened 2016-08-24
Moving (back) to marshalled session cookies

reidmix:
Currently with `:json` saved session cookies, the session data is transformed to json for saving which stringifies the keys of nested hashes.  Moving (back) to `:marshal` session cookies does not perform this transofrmation (I think it uses Ruby's marshalling methods which keeps the data's types). 

NOTE: Any test/development environments that have been logged into, you will need to clear your cookies (specifically the `_retrodot_session` cookie as they're not backwards compatible.

This came out of the Omniauth (#20) implementation.


lexelby:
Thanks again for figuring this out.  Google totally failed me.

lexelby:
:+1:

joshuatobin:
Whoa nice find. 👍

reidmix:
Yeah, it was totally a "find the right google incantation" to get to the answer.

PR #29, opened 2016-08-25
Less files for environment variables management

JuanitoFatas:
This Pull Request cleans up environment variable management. If this PR merged, one only needs to clone this PR and good to go! Ever need to override anything, put in `.env.local`. Test specific? Put in `.env.test`. :)

Ref. https://github.com/bkeepers/dotenv#multiple-rails-environments

<details>
  <summary>Details</summary>
- commit `.env` so faster developer setup
- ignore `.env.local`
- create failed environment variable in `.env` to tell something needs to
override, e.g.
    `INTERCOM_API_SECRET=intercom_api_secret`
- if needs to override, put in `.env.local`
- remove unnecessary files
- remove unnecessary lines in `spec/spec_helper.rb`
- `.env.{test,development}` is for environment-specific ENV
</details>


.env.test.sample
```diff
@@ -1,8 +0,0 @@
-SOURCE_URL="https://status.example.com/api/endpoint"
-DURATION="duration"
-STARTED_AT="started_at"
-INCIDENT_ID="incident_id"
-STATE="state"
-RESOLVED_AT="resolved_at"
-REVIEW="review"
-FOLLOWUP_DAYS="5"
```
joshuatobin: I'm not sure we want to be removing this file.

.gitignore
```diff
@@ -1,18 +1,13 @@
...
 /tmp
 /config/database.yml
-/.env
 /public/assets
-
 log
-
 *.swp
-.env
-.env.test
```
joshuatobin: Same here. I think we want to have a .gitignore for .env.test

reidmix: Agreed, and `.env`

.travis.yml
```diff
@@ -4,7 +4,6 @@ rvm:
 sudo: false
 bundler_args: --without development
 before_script:
-  - cp .env.test{.sample,} && cp .env{.sample,}
```
reidmix: I think we can keep this line too (maybe just remove the `.env` copy)

spec/spec_helper.rb
```diff
@@ -1,9 +1,6 @@
 ENV["RAILS_ENV"] ||= 'test'
 ENV["HOSTNAME"] = 'retrodot.test'
 
-require 'dotenv'
-Dotenv.load!('.env.test')
-
```
reidmix: Ditto, keep this.

reidmix:
I think there is confusion I can help with 💃 We use:
- `.env.development` is for when `RAILS_ENV=development` and is typically for when running your local server or pry, etc.
- `.env.test` is for `RAILS_ENV=test` and used by rspec and has test envvars needed for the specs to run 
- `.env` and `.env.local` is for all `RAILS_ENV`s

JuanitoFatas:
They have exact same contents:

- https://github.com/heroku/retrodot/blob/master/.env.test.sample
- https://github.com/heroku/retrodot/blob/master/.env.development.sample

So we should put into .env right? :)

.env will be loaded before the environment-specific env variants (.env.test, .env.development):

> When you load a certain environment, dotenv will first load general env vars from .env, then load environment specific env vars from .env.<current environment>. Variables defined in .env.<current environment> will override any values set in .env or already defined in the environment.
> -- https://github.com/bkeepers/dotenv#multiple-rails-environments

reidmix:
Good point.  Maybe?  @joshuatobin i know your usecase was that `.env` and `.env.test` were different and one leaking into the other?  I forget the specifics.

.env.test.sample
jmervine: @JuanitoFatas (cc @joshuatobin, @lexelby ) We tend to standardize on using `.env.test` to bootstrap our test configurations. As such, we'll probably want to keep this laying around. The one recommendation I can make is that we may want to forego having a `.env.test.sample` all together, in favor of just having `.env.test`. `.sample` indicates that we're expecting the developer to change something before using. In the case of this file, we'll want them to just use what's there in most cases.

.gitignore
jmervine: @JuanitoFatas (cc @joshuatobin, @lexelby, @reidmix) Per my comment above, we may want to leave `.env.test` off the ignore list. However, I strongly agree the `.env` should be in this file.

.travis.yml
jmervine: @JuanitoFatas (cc @joshuatobin, @lexelby, @reidmix) 👍 on this comment, also, once again, see my comment RE `.env.test` above.

JuanitoFatas:
Per https://github.com/heroku/retrodot/pull/29#discussion_r76827655, let's keep these envs as is. :) Thanks for the feedback!

joshuatobin:
@JuanitoFatas <3 - thanks for all the work thus far.

PR #30, opened 2016-08-25
Add batch update capability for categorizing incidents

joshuatobin:
This PR has the following changes:
* Adding a `batch_action` method for changing categories in ActiveAdmin
* Fixing a bug in the the Gemfile where webmock was being called in development
* Fixing a bug in authentication so that rails auto-reload works with ActiveAdmin
* Updates the associations so that `Incident`  belongs_to a `Category`.
* Spec for updating associations

CC: @heroku/sre


app/admin/incident.rb
```diff
@@ -15,15 +17,29 @@
     link_to 'Sync', sync_admin_incident_path(resource)
   end
 
+  # add a batch action for updating category
+  batch_action :category, form: ->{{category: Category.pluck(:name, :id)}} do |ids, inputs|
```
lexelby: The `batch_action :category` results in a menu item named "Category incidents".  Perhaps `batch_action :categorize`?

app/admin/incident.rb
```diff
@@ -15,15 +17,29 @@
     link_to 'Sync', sync_admin_incident_path(resource)
   end
 
+  # add a batch action for updating category
+  batch_action :categorize, form: ->{{category: Category.pluck(:name, :id)}} do |ids, inputs|
+    # inputs is a hash of all the form fields you requested
```
joshuatobin: @lexelby Fixed the action name ^

joshuatobin:
I've added some dummy data w/ categories for testing to the pr app.

lexelby:
I'll add a card to the board for making it possible to add categories via the admin UI.

lexelby:
This is freakin' awesome!!

lexelby:
:+1: from me.  A nice added touch might be to change the message to something like "Updated category for <number> incidents."

app/admin/incident.rb
```diff
@@ -15,15 +17,29 @@
...
+    ids.each do |id|
+      i = Incident.find(id)
+      i.category = Category.find(inputs["category"])
+      i.save!
+    end
+
+    redirect_to collection_path, notice: "Updated category for incidents!"
```
joshuatobin: @lexelby good call. Updated the messaging here.

PR #32, opened 2016-08-25
add config vars to app.json

lexelby:
This adds some of the settings from config.rb to app.json as optional params.  Adding them to app.json causes Heroku to inherit them from the template app (`retrodot-staging`) when creating a PR app.  We need this so that we can inherit heroku-specific custom variable values into our PR apps.

Setting `required = false` ensures that the PR app can be created without those values being set.

### DIFF
--- a/app.json
+++ b/app.json
@@ -25,7 +25,37 @@
     "SOURCE_URL": {
       "required": true
     },
-    "PR_APP": "true"
+
+    "PR_APP": "true",
+
+    "DURATION": {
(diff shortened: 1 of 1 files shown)


lexelby:
Details on PR app config value inheritance here: https://devcenter.heroku.com/articles/github-integration-review-apps#app-json

joshuatobin:
👍

lexelby:
Verified `retrodot-staging-pr-32` ended up with the config vars we want copied from staging.

PR #34, opened 2016-08-26
Adding License

joshuatobin:
CC: @heroku/sre


reidmix:
👍

LICENSE
```diff
@@ -0,0 +1,13 @@
+Copyright (c) 2016, Heroku
```
reidmix: I **think** keeping the "MIT License" at the top is the convention?

lexelby: I think so too.

PR #36, opened 2016-08-26
ChatOps api stub

lexelby:
This PR sets up a framework for the chatops API.  The goal is to produce an API that implements chatops commands, for consumption by a chatops plugin in your bot of choice (hubot, lita, etc).  Users can write a very small stub plugin for their bot that sends messages to Retrodot, receives a response from Retrodot, and posts it back to chat.

The retrodot chatops API has these two calls:

* `GET /api/chatops/v1/matcher`: get a regex that matches all chat commands retrodot cares about
* `POST /api/chatops/v1/respond`: process a message and return Retrodot's response

Both calls take an `API_KEY` GET param (can also be passed in the JSON POST body).  Must match the `CHATOPS_API_KEY` config var.

Requests/response format is not yet decided but will be in JSON.  

A secondary goal of this PR is to make the command processing reusable, in case we want to bring the chatops bot implementation into Retrodot (e.g. a slack /-command).  As such I've created a `ChatOps` module, and a `ChatOpsCommand` class that will have one child class for each command.  `ChatOpsCommand` registers child classes with `ChatOps`, which exposes methods to interact with the full set of chatops commands.  Client code only needs to run methods in `ChatOps` rather than instantiate individual command classes.  See `app/controllers/api/chatops/v1_controller.rb` for an example of how the `ChatOps` module is used.

I haven't yet figured out how I can write unit tests for this.  Advice appreciated!


app/chat_ops/chat_ops.rb
```diff
@@ -0,0 +1,41 @@
...
+
+      nil
+    end
+  end
+end
+
+# Load all .rb files in the 'commands' subdirectory using the require_all gem.
+# Rails's autoloading won't automatically load the subclasses, so without this
+# ChatOpsCommands.commands would return an empty array.
+require_rel 'commands'
```
reidmix: I might put this in the `lib` directory since it's "adjacent to" the app and supporting code for the app.

app/controllers/api/chatops/v1_controller.rb
```diff
@@ -0,0 +1,25 @@
...
+      render status: :not_found, json: {"error": "command not found"}
+    end
+  end
+
+  private
+    def check_api_key
+      unless params['API_KEY'] == Config.chatops_api_key
+        render status: :forbidden, json: {"error": "invalid api key"}
+      end
+    end
```
reidmix: Re: I suggest using basic auth / HTTP_AUTHORIZATION because it moves the authentication to a header and separate (well, a header) from the endpoint data, and is a common convention that we use (in watchtower, laika, API, etc.)  

Watchtower might be a good starting place, it's definitely not the prettiest code tho.

app/controllers/api/chatops/v1_controller.rb
```diff
@@ -0,0 +1,25 @@
+class Api::Chatops::V1Controller < ApplicationController
```
reidmix: I'd suggest putting the v1 bit after the API (ie. `Api::V1::Chatops` and `app/controllers/api/v1/v1_controller.rb`), again for convention, but does put everything "under" the v1 umbrella and the filename represents the Chatops class.

app/controllers/api/chatops/v1_controller.rb
reidmix: You could also create a `APIController` that everything descends from, where you can stash all the auth items in so you don't need to for subsequent endpoint controllers.

config/routes.rb
```diff
@@ -11,4 +11,11 @@
     get :verify
     get :failure
   end
+
+  namespace :api, defaults: {format: 'json'} do
```
reidmix: when you move v1, you can create a namespace for it here, too.

config/routes.rb
```diff
@@ -11,4 +11,11 @@
...
+    namespace :chatops do
```
reidmix: You can then create a `chatops` "resource"

app/chat_ops/chat_ops_command.rb
```diff
@@ -0,0 +1,31 @@
...
+
+  def process(user, message)
+    if result = self.class.regex.match(message)
+      run(user, result)
+    end
+  end
+
+  def run(user, message)
+    # define in subclass
+    {"error": "command not implemented"}
```
reidmix: would it be better to raise this as an error condition?

app/chat_ops/chat_ops.rb
```diff
@@ -0,0 +1,41 @@
...
+    # command matched.  TODO: define return value for success.
+    def process(user, message)
+      commands.each do |command|
+        if result = command.new.process(user, message)
+          return result
+        end
+      end
+
+      nil
+    end
```
reidmix: you can tighten this up using a `detect` and you won't need the `nil`

app/chat_ops/chat_ops.rb
lexelby: Heh.  I looked at a couple of articles and stackoverflow when I was deciding where to put this.  People said to only put stuff in `lib` if it would make sense to gemify it for use with multiple apps.  I can move it to `lib` if you still think that's where you want it, though?

app/chat_ops/chat_ops_command.rb
lexelby: Probably.  I'm going to nail down this whole protocol in a later PR though as I see what kinds of situations I end up needing to handle in the command classes.  Consider this a rough draft.  This way I don't make this whole thing as one gigantic PR :)

app/controllers/api/chatops/v1_controller.rb
lexelby: I had a hard time choosing this.  I decided to do it this way because I wanted to avoid having to increment versions for all of the APIs retrodot exposes at once.  Your way does have advantages too though.  I'll try it out and see how the code looks.

app/controllers/api/chatops/v1_controller.rb
lexelby: Ooh, good idea!

config/routes.rb
lexelby: Is that preferable even when I don't have a full set of CRUD operations?

app.json
```diff
@@ -55,6 +55,10 @@
     "REVIEW": {
       "description": "field name to use when syncing incidents from an external data source",
       "required": false
+    },
+
```
jmervine: Remove extra newline here.

lexelby: I was trying to break variables into logical sections as best I could given that JSON doesn't have comments.  Do you still think I should remove the newline?

app/chat_ops/chat_ops.rb
jmervine: I tend to agree with @reidmix here. The rule I follow is anything that doesn't logically fit the pattern for `app` (e.g. `model`, `mediator`, `notifier`, etc.) I put in `lib`. It keeps things cleaner.

jmervine:
@lexelby Looking good! Aside from @reidmix's comments, I'd probably rename files `s/chat_ops/chatops/`, but take that with a grain of salt, it's just a personal preference. =P

config/routes.rb
reidmix: oh maybe not... that's a good point.  `resource :chatops, include: []` seems silly.

lexelby:
@jmervine The problem with naming the directories/files `chatops` is that then the class/module ends up being `Chatops`, otherwise autoloading won't work.  `Chatops` looks wrong to me, honestly.  I actually have both, so I guess I should pick one and stick with it.

reidmix:
+1 on consistency.

lexelby:
Note that moving to `lib/` meant I had to add `lib/chat_ops` to the autoloader config explicitly, otherwise tests failed.

reidmix:
That's the usual way.  I tend to create a `lib/support` directory and let everything under there be autoloaded.

app/controllers/api/chatops/v1_controller.rb
lexelby: I'm gonna skip the `APIController` for now since the HTTP auth ended up being a one-line implementation.  I'll consider it when we add another API controller.

lexelby:
Okay, I think I've addressed all comments.  Ready for final +1s.

app.json
jmervine: Typically, I don't see spaces in JSON, however, as it's non-breaking, I'll leave it to you. =D

PR #37, opened 2016-08-31
Incident response models

lexelby:


spec/models/user_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'spec_helper'
+
+describe User do
```
JuanitoFatas: `RSpec.describe`?

spec/models/timeline_entry_spec.rb
```diff
@@ -0,0 +1,6 @@
+require 'spec_helper'
```
JuanitoFatas: [Already required](https://github.com/heroku/retrodot/blob/94b9baf149c7235b24ab0c76432f077357259ac9/.rspec#L2)?  :)

spec/models/user_spec.rb
reidmix: Agreed, I hope this isn't the new way.

db/migrate/20160830203407_create_incidents_users.rb
```diff
@@ -0,0 +1,8 @@
+class CreateIncidentsUsers < ActiveRecord::Migration[5.0]
+  def change
+    create_table :incidents_users, id: false do |t|
```
joshuatobin: `responders` maybe?

lexelby: Aren't you on vacation?

spec/models/user_spec.rb
lexelby: My bad, I typed this one by hand and must have just made an error that apparently works anyway.

spec/models/timeline_entry_spec.rb
lexelby: Neat! This was added by `rails generate`.  I'll try removing it.

spec/models/timeline_entry_spec.rb
```diff
@@ -0,0 +1,4 @@
+RSpec.describe TimelineEntry, type: :model do
```
JuanitoFatas: This `type: :model` can be removed because [this config](https://github.com/heroku/retrodot/blob/2c2380f17fb54dcce92e2487cdd285efb65ee869/spec/spec_helper.rb#L33) :)

lexelby: You are just a fount of useful information! Thanks :)

PR #23, opened 2016-08-23
Init a CONTRIBUTING.md [ci skip]

JuanitoFatas:
This Pull Request adds a contributing guidelines document to help others set up local dev environment and running tests.

[Preview CONTRIBUTING.md](https://github.com/JuanitoFatas/retrodot/blob/8270e6a/CONTRIBUTING.md).

PR #38, opened 2016-08-31
restore .env.sample removed in #23

lexelby:
#23 removed `.env.sample` (perhaps mistakenly?).  #29 also proposed to remove `.env.sample` and fixed `.travis.yml` to compensate.  #29 was since closed so Travis remains broken by #23 (it tries to copy `.env.sample` to `.env` and fails).

This PR restores `.env.sample` to fix travis.

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -1,0 +1,1 @@
+export SOURCE_URL="http://localhost"


lexelby:
@jmervine @joshuatobin @reidmix @JuanitoFatas

JuanitoFatas:
Sorry for the troubles here. I think someone mentioned the team actually uses `.env.development`, so I removed `.env.sample`.

Thanks for the fix :)

joshuatobin:
👍

reidmix:
That's correct, we moved to `.env.development`, Travis should be using `.env.test` so i'm not sure why it's broken other than `SOURCE_URL` not being defined in it the sample file?

reidmix:
I think the Travis script needs to accommodate not having a `.env.sample` was the problem (maybe?): 
`cp .env.test{.sample,} && cp .env{.sample,}`

lexelby:
Feel free to create a new PR -- I just needed to unblock myself :)

reidmix:
@lexelby would you mind fixing it if it hasn't been done already :D

lexelby:
@reidmix TBH, I'm so lost in all this .env / DotEnv business that I'd rather not create the PR.

reidmix:
@lexelby As an FYI, here's what's happening.  Before we used to rely on `.env` for development and `.env.test` for test.  But what happens is DotEnv loads in this order: `.env.local`, `.env.$RAILS_ENV`, `.env`

@joshuatobin was seeing that his `.env` variables were ending up in getting loaded (as expected by DotEnv) when running his tests.  We decided to use environmental environemnts correctly by using `.env.development` instead of just `.env`. 

Part of this change was forgetting to update `.travis.yml` to check for the existence of `.env` (which was moved to `.env.development`) before copying, which is the error you were seeing.

lexelby:
Gotcha, thanks

PR #39, opened 2016-08-31
Patch/cleanup specs

JuanitoFatas:
This PR cleans up specs for consistency :)

- enable [0 monkey patching mode](https://www.relishapp.com/rspec/rspec-core/v/3-5/docs/configuration/zero-monkey-patching-mode)

  top-level `describe` => `RSpec.describe`
- remove `require 'spec_helper'` ([already done in .rspec](https://github.com/heroku/retrodot/blob/2c2380f17fb54dcce92e2487cdd285efb65ee869/.rspec#L2))
- remove type metadata (already covered by [here](https://github.com/heroku/retrodot/blob/2c2380f17fb54dcce92e2487cdd285efb65ee869/spec/spec_helper.rb#L33))
- remove an outdated comment

### DIFF
--- a/spec/chat_ops/chat_ops_spec.rb
+++ b/spec/chat_ops/chat_ops_spec.rb
@@ -1,7 +1,3 @@
-require 'spec_helper'
-
-# Can't do `Rspec.describe ChatOps` here or the require_rel will get run before
-# we stub it below.
 RSpec.describe ChatOps do
   # This can't be a let because of weird ruby scoping stuff.
   REGEX1 = /test_regex1_[0-9]+/
--- a/spec/controllers/api/v1/chat_ops_controller_spec.rb
+++ b/spec/controllers/api/v1/chat_ops_controller_spec.rb
@@ -1,6 +1,6 @@
 require 'spec_helper'
 
-RSpec.describe Api::V1::ChatOpsController, type: :controller do
+RSpec.describe Api::V1::ChatOpsController do
   render_views
 
   describe "GET #matcher" do
--- a/spec/controllers/incidents_controller_spec.rb
+++ b/spec/controllers/incidents_controller_spec.rb
@@ -1,6 +1,4 @@
-require 'spec_helper'
-
-describe IncidentsController do
+RSpec.describe IncidentsController do
 
   let(:data) {
     {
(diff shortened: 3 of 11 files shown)


spec/controllers/api/v1/chat_ops_controller_spec.rb
```diff
@@ -1,6 +1,6 @@
 require 'spec_helper'
 
-RSpec.describe Api::V1::ChatOpsController, type: :controller do
+RSpec.describe Api::V1::ChatOpsController do
```
reidmix: Yes!

spec/chat_ops/chat_ops_spec.rb
```diff
@@ -1,7 +1,3 @@
-require 'spec_helper'
```
reidmix: So my mind is being blownup into a thousand peices -- do you know the details of what's loading `spec_helper` now?  Is that RSpec doing it (google isn't yielding me the answer)

JuanitoFatas: so .rspec already has [this command-line option](https://github.com/heroku/retrodot/blob/2c2380f17fb54dcce92e2487cdd285efb65ee869/.rspec#L2):

```
--require spec_helper
```

then every time when we run `rspec`, this command-line option will be included.

[.rspec file docs](https://www.relishapp.com/rspec/rspec-core/v/3-5/docs/configuration/read-command-line-configuration-options-from-files)

reidmix: Yup. The `--require` was the missing link for me.  I really like that solution, it DRYs it up nicely.

JuanitoFatas:
> on what changes I will need to make in how I write tests to be compatible with zero-monkey-patching mode. The doc you linked isn't particularly clear (to me).

Sorry wasn't being clear. This config only affects top-level `describe` per file, so make sure the first `describe` (or `feature`) in a spec file need to be `RSpec.describe` / `RSpec.feature`.

> I think you're saying that this:
> 
> x.should eq 2
> 
> becomes this:
> 
> expect(x).to eq 2

This is controlled by another config: [Syntax config](https://relishapp.com/rspec/rspec-expectations/docs/syntax-configuration). 😄

lexelby:
Got it, thanks!

PR #40, opened 2016-08-31
Improves ChatOps specs

JuanitoFatas:
This PR improves ChatOps specs.

- use `&:regex` because [it is faster](https://github.com/JuanitoFatas/fast-ruby#block-vs-symbolto_proc-code)

- move spec to a location that matched the class definition (`lib/chat_ops/chat_ops.rb` => `spec/lib/chat_ops/chat_ops.rb`)
  * Because some users may have editor plugin that automatically goes to the spec if the location matches, this help them find the correct spec
  * Also this consistency helps us find the right spec when our codebase grows
  
- (hopefully) improves specs
  * describe the method we are testing
  * decouples specs setup, no shared data => easily changed in the future
  * each spec follow this style: [setup exercise verify (teardown) style](https://robots.thoughtbot.com/four-phase-test)

    so from looking at each spec, we have a better idea of what being tested here.


spec/lib/chat_ops/chat_ops_spec.rb
```diff
@@ -0,0 +1,70 @@
+RSpec.describe ChatOps do
+  before do
+    ChatOps.class_variable_set :@@commands, []
```
lexelby: @reidmix This is like what we discussed, except I totally didn't know you could just reach in and set a class variable from outside a class!  Neato!  Sorta.  Heh.

@JuanitoFatas One issue here is that other tests might depend on the commands being correctly registered.  If they happen to run after this test then they will fail.  Can we save off a copy of `@@commands` here and restore it in `after`?

spec/lib/chat_ops/chat_ops_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+  describe '.process' do
+    let(:cmd_1_class) { Class.new(ChatOpsCommand) }
+    let(:cmd_2_class) { Class.new(ChatOpsCommand) }
+    let!(:cmd_1_instance) { cmd_1_class.new }
+    let!(:cmd_2_instance) { cmd_2_class.new }
+
+    before do
+      cmd_1_class.class_eval { match /test_regex1_[0-9]+/ }
+      cmd_2_class.class_eval { match /test_regex2_[a-z]+/ }
+      allow(cmd_1_class).to receive(:new) { cmd_1_instance }
```
lexelby: Is this preferable to using `expect_any_instance_of`?

spec/lib/chat_ops/chat_ops_spec.rb
JuanitoFatas: > If they happen to run after this test then they will fail.

This `before` block only affect tests within this file.

> Can we save off a copy of @@commands here and restore it in after?

Currently each test (`it`) is independent to each other, we can hold till the need comes :)

spec/lib/chat_ops/chat_ops_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+    end
+
+    it 'calls process on an instance of each defined command' do
+      expect(cmd_1_instance).to receive(:process)
+      expect(cmd_2_instance).to receive(:process)
+
+      ChatOps.process('user', 'message')
+    end
+
+    it 'calls run only on the command that matched' do
```
lexelby: We should assert that `cmd_1_instance` does not have its `run` method run.

spec/lib/chat_ops/chat_ops_spec.rb
JuanitoFatas: `expect_any_instance_of` is discouraged by RSpec itself:

> This feature is sometimes useful when working with legacy code, though in general we
discourage its use for a number of reasons:

> The rspec-mocks API is designed for individual object instances, but this feature operates on entire classes of objects. As a result there are some semantically confusing edge cases. For example, in expect_any_instance_of(Widget).to receive(:name).twice it isn't clear whether a specific instance is expected to receive name twice, or if two receives total are expected. (It's the former.)
Using this feature is often a design smell. It may be that your test is trying to do too much or that the object under test is too complex.
> It is the most complicated feature of rspec-mocks, and has historically received the most bug reports. (None of the core team actively use it, which doesn't help.)
> https://relishapp.com/rspec/rspec-mocks/docs/working-with-legacy-code/any-instance

And allow + expect we always have a clear idea what we are working on :)

spec/lib/chat_ops/chat_ops_spec.rb
JuanitoFatas: `cmd_1_class` doesn't have run method defined. So implicitly we are sure `run` method not run. But we can add another assertion here to make our intent more clear 👍  b7ba5e0

spec/lib/chat_ops/chat_ops_spec.rb
lexelby: Neat, thanks!

spec/lib/chat_ops/chat_ops_spec.rb
lexelby: [Sure it does](https://github.com/heroku/retrodot/blob/master/lib/chat_ops/chat_ops_command.rb#L27) :)

Thanks for adding the check.  If `cmd_1_instance.run` was called the test would have failed even without the extra assert, but since I plan to revamp how the return value / exception raising semantics of `process`/`run` work in the future, it's better to nail this down fully :)

spec/lib/chat_ops/chat_ops_spec.rb
lexelby: > This `before` block only affect tests within this file.

As far as I can tell, modifications to `ChatOps`'s `@@commands` last beyond each test.

We can hold off until later, but I'm concerned this may create a confusing gotcha for future test writers.  I like to make sure that no changes like this can escape the context of any test.

lexelby:
Thank you for rewriting my test, I knew it was kind of janky :D

lib/chat_ops/chat_ops.rb
```diff
@@ -18,7 +18,7 @@ def commands
     # Each ChatOpsCommand has its own regex, so we use Regexp.union to
     # combine them.  It's almost the same as joining them with '|'.
     def matcher
-      Regexp.union(commands.collect{|command| command.regex})
+      Regexp.union(commands.collect(&:regex))
```
lexelby: Ooh, it looks nicer too.

spec/lib/chat_ops/chat_ops_spec.rb
JuanitoFatas: > As far as I can tell, modifications to ChatOps's @@commands last beyond each test.

Indeed, that's why I clear `@@commands` before each test. So each test start fresh, not affected by others. :)


> but I'm concerned this may create a confusing gotcha for future test writers.

Same here, that's why this is at the very top. 

>  I like to make sure that no changes like this can escape the context of any test.

Hmm I cannot think of any case right now.

lexelby: Once I implement actual commands under lib/chat_ops/commands, I may want to test them through `ChatOps.process` using a unit test.  If that test runs before tests in this file, it will work, but if it runs after, `@@commands` will be empty and no code will be run.

JuanitoFatas: 👍 Please see 6ab3af2 :)

lexelby:
I'll merge once the conflicts are resolved.

JuanitoFatas:
> I'll merge once the conflicts are resolved.

resolved! 🎉

spec/lib/chat_ops/chat_ops_spec.rb
```diff
@@ -0,0 +1,72 @@
+RSpec.describe ChatOps do
+  let!(:commands) { ChatOps.commands }
+  before { ChatOps.class_variable_set :@@commands, [] }
+  after { ChatOps.class_variable_set :@@commands, commands }
```
reidmix: beautious

PR #41, opened 2016-08-31
Incident response library functions

lexelby:
This is me trying to speak like a rubyist :)

This PR will contain a couple of utility functions that will be used by chatops commands.  I had some fun with this, trying to get into the zen of thinking like a ruby programmer instead of just writing python in ruby.  I'd appreciate commentary on my use of iterators and other ruby idioms -- did I get it right?  Is it readable and pretty?

Still a WIP -- tests coming, along with a few other lib functions.


lexelby:
@jmervine @reidmix -- and @JuanitoFatas if you're interested :)

lib/incident_response.rb
```diff
@@ -0,0 +1,80 @@
...
+    def get_mentioned_users(message)
+      words = split_words(message)
+      users, words = search_for_names(words)
+      users += search_for_handles(words)
+      users.uniq
+    end
+
+    #private
+    def split_words(message)
+      message.scan(/[\w']+/).map &:downcase
```
jmervine: nit pick here, but stylistically would be `map(&:downcase)`, it's one of the few times where parens would be preferred.

lib/incident_response.rb
```diff
@@ -0,0 +1,80 @@
...
+    end
+
+    def get_users_by_name
+      Hash[User.all.collect{|user| [split_words(user.name), user]}]
```
jmervine: So if I'm reading this right, it would do something like ...

```ruby
users = [ 
    User.new(name: "John Doe", email: "john@doe.com", handle: "johnd"),
    User.new(name: "Jane Doe", email: "jane@doe.com", handle: "janed"),
]

get_users_by_name
#=> { 
#   [ [ "john", "doe" ], User(name: "John Doe", email: "john@doe.com", handle: "johnd") ] => nil,
#   [ [ "jane", "doe" ], User(name: "John Doe", email: "john@doe.com", handle: "janed") ] => nil, 
# } 
```

Or would it be...

```ruby
get_users_by_name
#=> { 
#   [ "john", "doe" ] => User(name: "John Doe", email: "john@doe.com", handle: "johnd") ],
#   [ "jane", "doe" ] => User(name: "John Doe", email: "john@doe.com", handle: "janed") ] 
# } 
```

Both seem incorrect.

jmervine: I **think** what you want here is something more like...

```ruby
def get_users_by_name
    users = {}
    User.all.each do |u|
       users[u.name.downcase] = u
    end

    users
end

# where...
get_users_by_name # with the above users
#=> { 
#  "john doe" => User(...), 
#  "jane doe" => User(...) 
# }
```

That said, there's probably an even better way of doing what you're trying to do through ActiveRecord, as any time you do `User.all`, it's going to be much slower than using `User.find_by`. Perhaps...

```ruby
# if `name_part` is the thing you're matching on...
User.where('lower(name) = ? or lower(handle) = ?', name_part.downcase).first
```

... something like that.

I could be totally misunderstanding your intentions though, or you could be using a ruby feature I'm not familiar with. =P

lexelby: It's this:

```
get_users_by_name
#=> { 
#   [ "john", "doe" ] => User(name: "John Doe", email: "john@doe.com", handle: "johnd"),
#   [ "jane", "doe" ] => User(name: "John Doe", email: "john@doe.com", handle: "janed") 
# } 
```

It creates a hash with tokenized names as keys and User objects as values.  In `search_for_names`, we're going to be pulling off sets of sequential words to see if they're a name, like this: `words[i,length]`.  Then it indexes the hash to see if that set of sequential words matches a name.

I'm pretty sure in this case using `User.all` and building a hash is vastly preferable to doing a bunch of queries.  I'd have to do one `User.where` for each n-gram in the message, which will quickly become a ton of queries especially if `Config.max_words_in_name` is 3 or more.

JuanitoFatas:
> and @JuanitoFatas if you're interested :)

@lexelby Yes I'm interested to help.

From what I see, we can give a constraint to incoming message (required mention by handles):

```
@juanitofatas and @lexelby will deploy this
```

Then implement our search function will be simpler:

```ruby
text = "@juanitofatas and @lexelby will deploy this"

handle_regexp = Regexp.new(User.pluck(:handle).join("|"), Regexp::IGNORECASE)

mentioned_handles = text.scan(handle_regexp)

users = User.where(handle: mentioned_handles)
```

What do you think? Does this match the requirements? Because I'm don't know what needs to be done here, can link me to the requirements or show some tests? 😊 

Also what's the format of handle? Can add validation + specs? 🙋

lexelby:
Tests added! :) I was actually writing them when you wrote that comment ;)

We're matching an existing chatops implementation in our internal hubot, which is closed source.  Our bot tracks the people that are responding to an incident and the interesting things that happen -- timeline entries.  I might do something like this:

```
h start incident
h add JuanitoFatas to incident
h timeline looks like API is busted, fred is looking at it
[... more timeline entries ...]
h end incident
```

You can see that I've explicitly added you to the incident, but I also mentioned that fred is doing something -- our bot notices this and adds him to the incident too.  The idea is for the bot to be as smart as possible and record as much relevant information as possible while requiring the least amount of cognitive load on incident responders as possible, since they're busy trying to fix stuff.

In the existing implementation, users can be mentioned with or without an @-sign.  This means you can mention someone into an incident without necessarily pinging them, depending on the chat system and their client's settings.  It also means that the person recording the timeline entries doesn't have to remember that they're adding people -- the system just notices that they're involved in the incident and adds them.

It's probably still possible to construct a regex to do this.  Perhaps we could even do it for full names.  But I'm a bit concerned the code would start to get confusing/unclear...

BTW, `RegExp.union` is useful. :)  Also we'd probably want to use `RegExp.escape` just in case.

@reidmix @jmervine thoughts on all this?

spec/models/user_spec.rb
```diff
@@ -1,3 +1,9 @@
 RSpec.describe User do
-  it { should have_many(:timeline_entries) }
+  #it { should have_many(:timeline_entries) }
+  it "should only allow valid handles" do
+    expect(FactoryGirl.build(:user, handle: "abcd123abc")).to be_valid
+    expect(FactoryGirl.build(:user, handle: "abcD123abc")).not_to be_valid
+    expect(FactoryGirl.build(:user, handle: "@abc123abc")).not_to be_valid
+    expect(FactoryGirl.build(:user, handle: "")).not_to be_valid
```
JuanitoFatas: These FactoryGirl prefixes can be removed. We [already included](https://github.com/heroku/retrodot/blob/master/spec/spec_helper.rb#L36) in our spec. :)

lib/incident_response.rb
jmervine: I see what you're saying about the queries, I wasn't thinking with a flood of messages. Caching the user list would definitely be preferable. 

Running `User.all` on each message would probably be worst then running `User.where(...)` though.

lexelby:
No longer a WIP, this is code complete now.

lib/incident_response.rb
```diff
@@ -80,10 +99,14 @@ def strip_contraction(word_or_words)
     def search_for_handles(words)
       # Search for people by their handle, with or without @ prepended.
 
-      words.collect do |word|
+      Array(words).collect do |word|
         word.sub! /^@/, ''
         User.find_by(handle: [word, strip_contraction(word)].uniq)
       end.reject &:nil?
```
jmervine: Stylistically `&:foo` should be in parens, so... `end.reject(&:nil?)`.

spec/models/user_spec.rb
```diff
@@ -1,3 +1,9 @@
 RSpec.describe User do
-  it { should have_many(:timeline_entries) }
+  #it { should have_many(:timeline_entries) }
```
JuanitoFatas: Why comment out this line❓

lib/incident_response.rb
```diff
@@ -0,0 +1,112 @@
...
+
+    private
+    WORD_RE = /[\w']+/
+
+    def split_words(message)
+      message.scan(/[\w']+/).map(&:downcase)
+    end
+
+    def get_users_by_name
+      Hash[User.all.collect{|user| [split_words(user.name), user]}]
```
JuanitoFatas: Instead of taking whole user name + user object out, consider using this:

```ruby
User.pluck(:id, :name) 
```

lib/incident_response.rb
```diff
@@ -0,0 +1,112 @@
+module IncidentResponse
+  class << self
+    # Find any users mentioned in the message and return them.  Users may be
+    # mentioned by handle (with or without @ prepended) or by their full name.
+    def get_mentioned_users(message)
+      words = split_words(message)
+      users, words = search_for_names(words)
+      users += search_for_handles(words)
+      users.uniq
```
JuanitoFatas: After https://github.com/heroku/retrodot/pull/41/files#r77247056, here should be a list of user id, then to return user objects:

```ruby
User.where(id: user_ids)
```

lexelby:
@JuanitoFatas 2e98f44 is for you. :)  I reworked a lot of the code to build regexes, both for user names and handles.  The biggest stumbling block was that it was tricky to go from the matched name (which is perhaps not in the right case, has excess whitespace, etc) back to the user.  I think the result is much simpler and more readable, though.  It might be faster too, but I'm not sure, with all those lookahead/lookbehind assertions.

lib/incident_response.rb
lexelby: Ah, I see.  I'm used to ORMs that would optimize taking the User object out by lazy-loading its contents.  Fixed in f4ee0ab.

spec/models/user_spec.rb
lexelby: Whoops.

lib/incident_response.rb
lexelby: I'm not talking about a flood of messages.  The old algorithm worked by considering sets of consecutive words.  For example, for this:

> This is a six-word sentence.

it would look for names like this:

1. this is a
2. is a six
3. a six word
4. six word sentence
5. this is
6. is a
...

If we called `User.find_by` every time, we'd be doing a _lot_ of queries for each message.  I would argue that the number of queries (and round-trips to the DB) would outweigh pulling the entire User table, especially if we use `pluck` as Juanito suggests.

I've since reduced `get_mentioned_users` to four queries.  I could potentially reduce it to three if I plucked name, handle, and id all at once, but I think that would probably hurt the code structure too much for not enough gain.

Also note that I can't cache the user list between messages because users may be added at any time.  I'm not running this code on every single message sent in chat, only timeline entries (`h timeline <message>`), so I think this is fine.

lib/incident_response.rb
```diff
@@ -58,7 +61,7 @@ def search_for_names(message)
     end
 
     def handles_regex
-      handles = User.pluck(:handle)
+      handles = User.pluck(:handle).reject(&:nil?).reject(&:empty?)
```
JuanitoFatas: Or `reject(&:blank?)` :)

lib/incident_response.rb
```diff
@@ -0,0 +1,80 @@
...
+      [mentions, message]
+    end
+
+    def handles_regex
+      handles = User.pluck(:handle).reject(&:nil?).reject(&:empty?)
+      /(?<=^|\W)@?(#{Regexp.union(handles)})(?=\W|$)/i
+    end
+
+    def search_for_handles(message)
+      # Search for people by their handle, with or without @ prepended.
```
JuanitoFatas: How about put this comment above its definition (above `def search_for_handles(message)`? Ruby does not have Python's powerful docstring :'(

lib/incident_response.rb
```diff
@@ -0,0 +1,80 @@
...
+        # matches on users' handles in search_for_handles.
+        " "
+      end
+
+      [mentions, message]
+    end
+
+    def handles_regex
+      handles = User.pluck(:handle).reject(&:nil?).reject(&:empty?)
+      /(?<=^|\W)@?(#{Regexp.union(handles)})(?=\W|$)/i
```
JuanitoFatas: `/@?(#{Regexp.union(handles)})/i` also works? How about adding some tests to demonstrate the use of `(?<=^|\W)` and `(?=\W|$)` so that future us won't break it. :)

JuanitoFatas:
> In short: because there is no Config.max_words_in_name anymore (oops, need to remove this from config.rb, will do shortly).

Got it  👍

lib/incident_response.rb
```diff
@@ -0,0 +1,80 @@
...
+      names = names_to_users.keys.map do |name|
+        name.split.map{|word| Regexp.escape(word)}.join('\s+')
+      end
+
+      # Match longer names first.  This makes sure that "John Smith-Jones"
+      # doesn't match "John Smith".
+      names = names.sort_by {|name| -name.length}
+
+      names.reject!(&:nil?)
+      names.reject!(&:empty?)
```
joshuatobin: Probably doesn't matter either way but I think this could be `names.reject!(&:nil?).reject!(&:empty?)` ?

lib/incident_response.rb
```diff
@@ -0,0 +1,80 @@
...
+
+    def search_for_names(message)
+      names_to_users = {}
+
+      User.pluck(:name, :id).each do |name, id|
+        names_to_users[normalize_name(name)] = id
+      end
+
+      # Build an array of regexes, one per name.  We need to match without
+      # regard for whitespace, so convert spaces between words into '\s+'.
```
joshuatobin: The second sentence reads a bit strange. `We need to match without regard for ...` ?

spec/factories/user.rb
```diff
@@ -0,0 +1,7 @@
+FactoryGirl.define do
+  factory :user do
+    sequence(:name) {|n| "User#{n} Lastname" }
+    sequence(:handle) {|n| "user#{n}" }
+    sequence(:email) {|n| "user#{n}@example.com" }
```
joshuatobin: nitpick. Its much easier to read if the blocks are lined up.

lib/incident_response.rb
```diff
@@ -0,0 +1,80 @@
...
+
+      message.gsub! names_regex do |match|
+        mentions << names_to_users[normalize_name(match)]
+
+        # Remove the name from the message so that it is not considered for
+        # matches on users' handles in search_for_handles.
+        " "
+      end
+
+      [mentions, message]
```
joshuatobin: This feel somewhat python-ish to me. I'd need to play with it a bit but I think you might be able to just use `map!` or possibly `each_with_object([])`?

lib/incident_response.rb
lexelby: Here you go: efcb664b8f.  Try removing the lookahead and lookbehind assertions and see what happens :)

lib/incident_response.rb
lexelby: Oh, sure enough!  Thanks.

lexelby: Turns out that we can't do this.  If `reject!` removes items from the array, it returns the array.  But if it makes no changes, it returns `nil`.  That's kind of a stupid behavior in my mind.

lib/incident_response.rb
lexelby: Hmm, how about: "We should match a name even if the number of spaces between the words is different ..."?

lib/incident_response.rb
lexelby: Oh, interesting.  Perhaps the method should be `find_and_remove_names!`?

lib/incident_response.rb
joshuatobin: 👍

lib/incident_response.rb
```diff
@@ -0,0 +1,80 @@
...
+
+      # Build an array of regexes, one per name.  We need to match without
+      # regard for whitespace, so convert spaces between words into '\s+'.
+      names = names_to_users.keys.map do |name|
+        name.split.map{|word| Regexp.escape(word)}.join('\s+')
+      end
+
+      # Match longer names first.  This makes sure that "John Smith-Jones"
+      # doesn't match "John Smith".
+      names = names.sort_by{|name| name.length}.reverse
```
joshuatobin: 👍 <3

joshuatobin:
👍

lib/incident_response.rb
```diff
@@ -0,0 +1,75 @@
...
+
+    private
+    def normalize_name(name)
+      name.downcase.strip.gsub(/\s+/, ' ')
+    end
+
+    def find_and_remove_names!(message)
+      names_to_users = User.pluck(:name, :id).each_with_object({}) do |(name, id), h|
+        h[normalize_name(name)] = id
+      end
```
joshuatobin: @lexelby thanks <3. This is very nice.

lib/incident_response.rb
```diff
@@ -0,0 +1,75 @@
...
+    def prevent_highlights(message)
+      message.gsub(handles_regex) do |handle|
+        unhighlight(handle)
+      end
+    end
+
+    private
+    def normalize_name(name)
+      name.downcase.strip.gsub(/\s+/, ' ')
+    end
```
reidmix: Any way to normalize the names as they're entered into the db? So this would happen on entry once, and not after that.

lib/incident_response.rb
reidmix: It makes me very happy.

reidmix: FWIW, you could do this too if you could code `normalize_name` away:
```ruby
Hash[User.pluck(:name, :id)]
```

lib/incident_response.rb
reidmix: You can also: `names.sort_by(&:length).reverse`

lib/incident_response.rb
```diff
@@ -0,0 +1,75 @@
...
+      names = names_to_users.keys.map do |name|
+        name.split.map{|word| Regexp.escape(word)}.join('\s+')
+      end
+
+      # Match longer names first.  This makes sure that "John Smith-Jones"
+      # doesn't match "John Smith".
+      names = names.sort_by{|name| name.length}.reverse
+
+      names.reject!(&:nil?)
+      names.reject!(&:empty?)
```
reidmix: Q: Why would a name be nil/empty -- meaning, can we prevent these from coming in (at either sync time or when you perform `User.pluck`) ?

reidmix: FWIW, `blank?` will cover both the `nil?` and `empty?` case -- it's a Rails addition.

lib/incident_response.rb
reidmix: same notes about nil/empty as in `find_and_remove_names` -- it might make it simpler to prevent those coming in with a scope, something like `User.with_handles.pluck(:handle)`

lib/incident_response.rb
```diff
@@ -0,0 +1,75 @@
...
+      handles = find_handles(message)
+
+      (User.where(id: ids) + User.where(handle: handles)).uniq
+    end
+
+    # Slack doesn't nominally have a way to prevent a mention of a user's
+    # handle from triggering highlights.  We intersperse each character of
+    # the handle with the unicode character "invisible separator" which has
+    # zero width.  The result is visually indistinguishable from their name
+    # but doesn't trigger highlights.
```
reidmix: Oh Slack! :/

spec/factories/user.rb
```diff
@@ -0,0 +1,7 @@
+FactoryGirl.define do
+  factory :user do
+    sequence(:name)   {|n| "User#{n} Lastname" }
```
reidmix: There's a cool library called [faker](https://github.com/stympy/faker) that will generate names, emails, etc. for you which is super convenient and more real-world.

lib/incident_response.rb
```diff
@@ -0,0 +1,75 @@
...
+      message.scan(handles_regex).map(&:first).map do |handle|
+        handle.downcase
+      end
+    end
+
+    def unhighlight(handle)
+      handle.each_char.to_a.join("\u{2063}")
+    end
+  end
+end
```
reidmix: @lexelby I really like you're approach here.

I was thinking about pulling from the db every user to make matches (something you could cache on later if you needed), versus doing the opposite and how that's better than scanning the handles for each word which clearly wouldn't be the way to go.

spec/lib/incident_response_spec.rb
```diff
@@ -0,0 +1,92 @@
+RSpec.describe IncidentResponse do
```
reidmix: I get it now -- this prevents polluting the global method namespace with a method called `describe`, but once you're in the block, the DSL takes over.  I'm liking this once I did a TIL.

spec/lib/incident_response_spec.rb
```diff
@@ -0,0 +1,92 @@
...
+      expect(IncidentResponse.prevent_highlights(before)).to eq after
+    end
+
+    it "should work even for contractions" do
+      before = "lorem ipsum foo're dolor sit amet"
+      after = "lorem ipsum f#{separator}o#{separator}o're dolor sit amet"
+      expect(IncidentResponse.prevent_highlights(before)).to eq after
+    end
+  end
+end
```
reidmix: These are great tests!

lexelby: Thanks!  They've already caught a few bugs :D

reidmix:
Ping @lexelby there were a few comments on here that may have slipped away (or solved in other PRs?)

lexelby:
@reidmix I think everything is resolved?  Did you have specific comments you thought hadn't been addressed yet?

lib/incident_response.rb
reidmix: Curious if this can get normalized before db insert? ^

lib/incident_response.rb
reidmix: If it can get normalized, you can shorten this ^

lib/incident_response.rb
reidmix: A little `to_proc` syntax sugar ^

lib/incident_response.rb
reidmix: Suggestion to retrieve only users with handles, then you don't have to reject ^ use a scope to make it simpler ^

spec/factories/user.rb
reidmix: Suggestion to use `faker` to generate names, handles, and emails ^

lib/incident_response.rb
lexelby: Nope, definitely not a good idea.  They're people's names, so we can't make assumptions about capitalization, etc.  About the best we could do would be to strip leading and trailing whitespace, but I don't think that would help this code much.

spec/factories/user.rb
lexelby: Useful, but I don't think it's important enough to warrant opening a new PR at this point.  I'll keep it in mind for the future.

PR #42, opened 2016-09-02
add sync:users rake task

lexelby:
As a temporary measure for our MVP, we're piggybacking off an existing chatops bot rather than implementing a new bot in Retrodot.  This PR syncs retrodot's Users with a list fetched from the bot.

Tests coming soon.


Gemfile
```diff
@@ -27,6 +27,7 @@ gem 'inherited_resources', github: 'activeadmin/inherited_resources'
 gem 'omniauth-google-oauth2'
 gem 'rack-ssl-enforcer'
 gem 'require_all'
+gem 'active_record_ignored_attributes'
```
joshuatobin: Neat! I like this gem.

app/models/user.rb
```diff
@@ -1,5 +1,13 @@
...
-  validates :handle, format: { with: /\A[a-z0-9]+\z/, message: "only allows numbers and lowercase letters" }
+  validates :handle, format: { with: /\A[-_.a-z0-9]+\z/, message: "only allows numbers, lowercase letters, dashes, periods and underscores" }
+
+  def as_json(options={})
+    {
+      name: name,
+      email: email,
+      handle: handle
+    }
+  end
```
joshuatobin: I believe you can get away from doing this with `active_model_serializers` gem?

joshuatobin: Also since this is an MVP I'd say its fine for now, just mentioning the above for future.

joshuatobin:
👍

Gemfile
lexelby: Me too, all except for the name ;)

app/models/user.rb
lexelby: Yeah, I kind of suspected as much but I didn't want to learn `active_model_serializers` :)

Gemfile
reidmix: Wow, so good.

app/models/user.rb
```diff
@@ -1,5 +1,13 @@
 class User < ApplicationRecord
   has_and_belongs_to_many :incidents, join_table: :incidents_responders
   has_many :timeline_entries
-  validates :handle, format: { with: /\A[a-z0-9]+\z/, message: "only allows numbers and lowercase letters" }
+  validates :handle, format: { with: /\A[-_.a-z0-9]+\z/, message: "only allows numbers, lowercase letters, dashes, periods and underscores" }
```
reidmix: A suggestion on messages/copy, to start using the `config/locale/en.yml` dictionary, it helps remove much of this copy to a standard place and keeps the code somewhat more concise

app/models/user.rb
reidmix: FWIW, they're fairly easy.

app/support/mediators/user/sync.rb
```diff
@@ -0,0 +1,24 @@
...
+    def call
+      users.each do |chatops_user|
+        if chatops_user['email']
+          user = ::User.find_or_initialize_by(email: chatops_user['email'])
+          user.name = chatops_user['name'] || ""
+          user.handle = chatops_user['handle']
+          user.save!
+        end
+      end
+    end
```
reidmix: on `Mediators::Base` you could run `with_indifferent_access` on options when saving them, then use symbols as keys here.  My preference is using symbols in cases like this because it's easier to type and keep it consistent with other parts of the apps where they're using symbols.

app/support/mediators/user/sync.rb
```diff
@@ -0,0 +1,24 @@
+module Mediators::User
+  class Sync < Mediators::Base
+    def call
+      users.each do |chatops_user|
+        if chatops_user['email']
+          user = ::User.find_or_initialize_by(email: chatops_user['email'])
+          user.name = chatops_user['name'] || ""
```
reidmix: it might be worth to keep it `nil` for simplicity (in creating queries and the user lookup methods in #41)

app/support/mediators/user/sync.rb
```diff
@@ -0,0 +1,24 @@
...
+      end
+    end
+
+    private
+    def users
+      excon = Excon.new(Config.chatops_users_url)
+      result = excon.request(method: :get,
+                             query: {secret: Config.chatops_users_api_key},
+                             expects: 200,
+                             idempotent: true)
```
reidmix: You can combine these two:
```ruby
result = Excon.new(url, method: :get, ...)
MultiJson.decode(result.body)
```

spec/support/mediators/user/sync_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+
+    let(:modified_user1) { build(:user, name: "changed name", handle: user1.handle, email: user1.email) }
+
+    it "loads users returned from the chatops users request" do
+      stub_request(:get, url).to_return body: [user1.as_json, user2.as_json].to_json
+
+      Mediators::User::Sync.run
+
+      expect(User.find_by(email: user1.email)).to be_same_as user1
+      expect(User.find_by(email: user2.email)).to be_same_as user2
```
reidmix: nice use of `same_as?` 👍

lexelby: Thanks. :)  I knew there had to be a gem for this.

reidmix:
Ping @lexelby there were a few comments on here that may have slipped away.

app/models/user.rb
lexelby: We need to fix this validator to allow nil anyway, so I'll fix up the messages then too.

app/support/mediators/user/sync.rb
lexelby: Cool, I'll keep that in mind.  In this case, the hashes aren't from `options`, they're from a JSON response.  I'm not sure how to `with_indifferent_access` or `symbolize_keys` recursively through a data structure...

app/support/mediators/user/sync.rb
lexelby: Fixed in a later PR (made `User.name` nullable)

app/support/mediators/user/sync.rb
lexelby: Ah, thanks.  Looks like this one slipped through and I don't feel it's worth an entire PR to fix it, but I'll keep this in mind.

app/support/mediators/user/sync.rb
reidmix: `with_indifferent_access` doesn't actually change the underlying structure but does work on embedded hashes:
```ruby
{"a" => {"b" => "c"}}.with_indifferent_access[:a][:b] #=> "c"
```

reidmix: Looks like this is resolved in future PRs too by using the `slice`/`symobolize_keys` methods 👍

PR #44, opened 2016-09-07
"start incident" command

lexelby:
Posting a WIP because folks were curious about my time zone woes.

The highlight reel:

* The old chatops stuff could handle '15:12 EDT', for example, along with natural language like "15 minutes ago".
* Ruby's canonical gem for natural language date parsing is Chronic.
* Chronic pukes on "15:12 EDT" and returns `nil`.
* I assumed this was because it couldn't parse time zones.
* I spent a very very long time trying to figure out how to parse "EDT", "PDT", "PST", etc in rails.
  * Spoiler: it's really difficult.  Indiana makes life very difficult.
* Turns out that Chronic only hates time zones [sometimes](https://github.com/mojombo/chronic/issues/134), so there's a relatively simple workaround.
* denouement: I can never remember if it's "EDT" or "EST" at any given time of the year, so I made the code pick the correct one no matter which the user says.

cc @jmervine @reidmix

### DIFF
--- a/app/controllers/api/v1/chat_ops_controller.rb
+++ b/app/controllers/api/v1/chat_ops_controller.rb
@@ -7,7 +7,7 @@ class Api::V1::ChatOpsController < ApplicationController
   end
 
   def respond
-    user = User.ensure(**params['user'].slice('email', 'name', 'handle').symbolize_keys)
+    user = User.ensure(**params.permit(user: [:email, :name, :handle])[:user].to_h.symbolize_keys)
     result = ChatOps.process(user, params['message'])
 
     if result
--- a/app/controllers/api/v1/chat_ops_controller.rb
+++ b/app/controllers/api/v1/chat_ops_controller.rb
@@ -7,7 +7,8 @@ class Api::V1::ChatOpsController < ApplicationController
   end
 
   def respond
-    result = ChatOps.process(params['user'], params['message'])
+    user = User.ensure(**params['user'].slice('email', 'name', 'handle').symbolize_keys)
+    result = ChatOps.process(user, params['message'])
 
     if result
       render json: result
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -10,4 +10,13 @@ class User < ApplicationRecord
       handle: handle
     }
   end
+
+  class << self
+    def ensure(email:, name: nil, handle: nil)
+      user = find_or_create_by(email: email)
+      user.name = name
+      user.handle = handle
+      user.save!
+    end
+  end
 end
--- a/app/support/mediators/user/sync.rb
+++ b/app/support/mediators/user/sync.rb
@@ -3,10 +3,7 @@ module Mediators::User
     def call
       users.each do |chatops_user|
         if chatops_user['email']
-          user = ::User.find_or_initialize_by(email: chatops_user['email'])
-          user.name = chatops_user['name'] || ""
-          user.handle = chatops_user['handle']
-          user.save!
+          User.ensure(**chatops_user.slice('email', 'name', 'handle').symbolize_keys)
         end
       end
     end
--- a/db/migrate/20160908150249_change_users.rb
+++ b/db/migrate/20160908150249_change_users.rb
@@ -1,0 +1,6 @@
+class ChangeUsers < ActiveRecord::Migration[5.0]
+  def change
+    change_column_null(:users, :name, true)
+    add_index :users, :email, unique: true
+  end
+end
--- a/lib/chat_ops/chat_ops.rb
+++ b/lib/chat_ops/chat_ops.rb
@@ -21,6 +21,11 @@ module ChatOps
       Regexp.union(commands.collect(&:regex))
     end
 
+    # Get a help message describing each command.
+    def help
+      commands.map(&:help).reject(&:blank?).join("\n")
+    end
+
     # Try to process a message as a ChatOps command.  Return nil if no
     # command matched.  TODO: define return value for success.
     def process(user, message)
--- a/lib/chat_ops/commands/start_incident_command.rb
+++ b/lib/chat_ops/commands/start_incident_command.rb
@@ -1,76 +1,77 @@
-class StartIncidentCommand < ChatOpsCommand
-  match %r{
-          # allow "start incident" or "start an incident"
-          start\s+(an\s+)?incident
-          (\s+
-            (
-              # if they specify an incident ID, use that.
-              (?<incident_id>\d+)
+module ChatOps::Commands
+  class StartIncidentCommand < ChatOpsCommand
+    match %r{
+            # allow "start incident" or "start an incident"
+            start\s+(an\s+)?incident
+            (\s+
+              (
+                # if they specify an incident ID, use that.
+                (?<incident_id>\d+)
 
-              # incident ID must be followed by whitespace
-              \s+
+                # incident ID must be followed by whitespace or EOL
+                (\s+|$)
 
-              # Don't match incident #14 in "start an incident 14 minutes ago"
-              (?!(seconds|minutes|hours)\s+ago)
+                # Don't match incident #14 in "start an incident 14 minutes ago"
+                (?!(seconds|minutes|hours)\s+ago)
 
+              )?
+
+              # Slurp up the remainder.  IncidentResponse.parse_timestamp will
+              # parse it as a timestamp specified in natural language.
+              (?<timestamp>.*)
             )?
+            $
+          }ix
 
-            # Slurp up the remainder.  IncidentResponse.parse_timestamp will
-            # parse it as a timestamp specified in natural language.
-            (?<timestamp>.*)
-          )?
-          $
-        }ix
+    help_message "h start incident [#] [at <timespec>] - sets or overwrites incident start time (timespec examples: 5 minutes ago, 3pm, etc)"
 
-  help_message "h start incident [#] [at <timespec>] - sets or overwrites incident start time (timespec examples: 5 minutes ago, 3pm, etc)"
+    def run(user, match)
+      incident_id = infer_incident_id(match[:incident_id])
+      chat_start = get_chat_start(match[:timestamp]).in_time_zone(Config.time_zone)
 
-  def run(user, match)
-    incident_id = infer_incident_id(match[:incident_id])
-    chat_start = get_chat_start(match[:timestamp]).in_time_zone(Config.time_zone)
+      message = ["Recorded the start of chat for incident \##{incident_id} at #{chat_start.inspect}"]
 
-    message = ["Recorded the start of chat for incident \##{incident_id} at #{chat_start.inspect}"]
+      incident = Incident.find_or_create_by(incident_id: incident_id)
 
-    incident = Incident.find_or_create_by(incident_id: incident_id)
+      if incident.chat_start
+        message << "     (overwriting start time for incident \##{incident_id}, was: #{incident.chat_start.inspect})"
+      end
 
-    if incident.chat_start
-      message << "     (overwriting start time for incident \##{incident_id}, was: #{incident.chat_start.inspect})"
+      incident.timeline_start = Time.now
+      incident.chat_start = get_chat_start(match[:timestamp])
+      incident.responders << user
+      incident.save
+
+      ChatOps.help.each_line do |line|
+        message << "  " + line
+      end
+
+      { message: message.join("\n") }
     end
 
-    incident.timeline_start = Time.now
-    incident.chat_start = get_chat_start(match[:timestamp])
-    incident.responders << user
-    incident.save
+    private
+    def infer_incident_id(incident_id=nil)
+      # If they specified an incident ID, use it.
+      # Otherwise, if an incident is open, they probably mean that one.
+      incident_id = Incident.open.first&.incident_id if !incident_id
 
-    ChatOps.help.each_line do |line|
-      message << "  " + line
+      # If no incident is open, they probably mean the ID of the next incident to
+      # be opened.
+      begin
+        incident_id = Incident.synced.first.incident_id + 1 if !incident_id
+      rescue NoMethodError
+        # If no incidents have been opened yet, we would have tried to do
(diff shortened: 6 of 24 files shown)


lib/incident_response.rb
```diff
@@ -20,7 +20,58 @@ def prevent_highlights(message)
...
+
+        return result
+      end
+    end
+
     private
+    # If passed "EST" during daylight time, return "EDT", and the reverse.
+    # People often say "3pm EST" when it's technically "3pm EDT" during that
+    # part of the year.
+    def fix_dst(tz)
```
joshuatobin: <3. I always do this.

lexelby: Hah, excellent.  I was going a bit out on a limb when I said "People often say..." when I only had myself as a data point, but I had a hunch ;)

lib/incident_response.rb
```diff
@@ -20,7 +20,58 @@ def prevent_highlights(message)
       end
     end
 
+    # Parse a timestamp specified in natural language.
+    def parse_timestamp(timestamp)
+      Time.use_zone(Config.time_zone) do
+        Chronic.time_class = Time.zone
+
+        # Chronic likes "today at 3pm" but not "at 3pm".
+        timestamp.sub! /^at /, ''
```
joshuatobin: wow that is super annoying :/ Wonder if its worth shooting them a PR.

lexelby: Nope.  The project is largely unmaintained.

joshuatobin:
@lexelby is this still a WIP ?

lexelby:
Yup.  Should have a non-WIP pushed up within the hour.

lexelby:
Okay, I'm repurposing this PR to be just "start incident" and the associated plumbing (natural language date parsing) because it's starting to get big.  This should be good to go, once I merge in master.

lexelby:
@tobin @reidmix this one's ready for review

app/support/mediators/user/sync.rb
```diff
@@ -3,10 +3,7 @@ class Sync < Mediators::Base
     def call
       users.each do |chatops_user|
         if chatops_user['email']
-          user = ::User.find_or_initialize_by(email: chatops_user['email'])
-          user.name = chatops_user['name'] || ""
-          user.handle = chatops_user['handle']
-          user.save!
+          User.ensure(**chatops_user.slice('email', 'name', 'handle').symbolize_keys)
```
joshuatobin: I <3 this :D

db/migrate/20160908150249_change_users.rb
```diff
@@ -0,0 +1,6 @@
+class ChangeUsers < ActiveRecord::Migration[5.0]
+  def change
+    change_column_null(:users, :name, true)
```
joshuatobin: Why are we allowing `name` to be null?

lib/chat_ops/chat_ops.rb
```diff
@@ -21,6 +21,11 @@ def matcher
       Regexp.union(commands.collect(&:regex))
     end
 
+    # Get a help message describing each command.
+    def help
+      commands.map(&:help).reject(&:blank?).join("\n")
```
joshuatobin: <#

db/migrate/20160908150249_change_users.rb
lexelby: There's no reason to require a name, and it was making the code complicated and weird to have name be required.  The unique identifier is email.

joshuatobin: We answer this in pairing, but for the record, its because we really only care about email

joshuatobin:
👍

lib/chat_ops/commands/start_incident_command.rb
```diff
@@ -1,3 +1,77 @@
...
+                (?!(seconds|minutes|hours)\s+ago)
+
+              )?
+
+              # Slurp up the remainder.  IncidentResponse.parse_timestamp will
+              # parse it as a timestamp specified in natural language.
+              (?<timestamp>.*)
+            )?
+            $
+          }ix
```
reidmix: Lovely

lib/chat_ops/commands/start_incident_command.rb
```diff
@@ -1,3 +1,77 @@
...
+    def run(user, match)
+      incident_id = infer_incident_id(match[:incident_id])
+      chat_start = get_chat_start(match[:timestamp]).in_time_zone(Config.time_zone)
+
+      message = ["Recorded the start of chat for incident \##{incident_id} at #{chat_start.inspect}"]
+
+      incident = Incident.find_or_create_by(incident_id: incident_id)
+
+      if incident.chat_start
+        message << "     (overwriting start time for incident \##{incident_id}, was: #{incident.chat_start.inspect})"
```
reidmix: you could use `incident.chat_start.rfc822` or `incident.chat_start.to_s(:rfc822)` (although they have the timezone offset, they don't have the timezone name).  Suggesting this because inspect is usually for debugging, but here it's funny because it gives a human readable representation!

lib/chat_ops/commands/start_incident_command.rb
```diff
@@ -1,3 +1,77 @@
...
+      incident.chat_start = get_chat_start(match[:timestamp])
+      incident.responders << user
+      incident.save
+
+      ChatOps.help.each_line do |line|
+        message << "  " + line
+      end
+
+      { message: message.join("\n") }
+    end
```
reidmix: My general thought around this method is that I'd want to separate out the action (here, saving the incident) from the messaging.  For me, having them together muddies up the method.  

I think the primary way to do that is move `match` into an instance variable/getter (probably the same with `user`, `incident`) because those will always be available to a command.  Once you do that you'll be able to build decoupled methods so that run could look like:

```ruby
def run
  incident = Incident.find_or_create_by(incident_id: incident_id)
  incident.timeline_start = Time.now
  incident.chat_start = chat_start
  incident.responders << user
  incident.save

  # return message code (I see it changed later on)
end
```

Actually, because everything descends from `ChatOpsCommand` and it does the `process`ing.
You could have it retrieve the message on its own like so:
```ruby
def process(user, message)
  # if etc...
    run
    message
  #end
end
```
Assuming you have a `message` method that does the result messaging.

reidmix: This would also make it eaiser to test both the `run`ing and the `message`ing more simply.

lib/chat_ops/commands/start_incident_command.rb
```diff
@@ -1,3 +1,77 @@
...
+
+    private
+    def infer_incident_id(incident_id=nil)
+      # If they specified an incident ID, use it.
+      # Otherwise, if an incident is open, they probably mean that one.
+      incident_id = Incident.open.first&.incident_id if !incident_id
```
reidmix: `unless` could work here.

reidmix: Actually, I'd put `return unless incident_id` right at the top of the method.

lib/chat_ops/commands/start_incident_command.rb
reidmix: I know that I'm reviewing this at towards the end and not the beginning where this would have been more useful. This is a suggestion, and probably one that you can save until after MVP when you can do a refactor .

Assuming this all makes sense, I can talk you through my thinking if not.

lib/chat_ops/commands/start_incident_command.rb
reidmix: `incident_id ||= Incident...` will tidy this up.

lib/chat_ops/commands/start_incident_command.rb
```diff
@@ -1,3 +1,77 @@
...
+
+      # If no incident is open, they probably mean the ID of the next incident to
+      # be opened.
+      begin
+        incident_id = Incident.synced.first.incident_id + 1 if !incident_id
```
reidmix: incident_id ||= Incident... will tidy this up.

lib/chat_ops/commands/start_incident_command.rb
```diff
@@ -1,3 +1,77 @@
...
+      rescue NoMethodError
+        # If no incidents have been opened yet, we would have tried to do
+        # nil.incident_id, which would raise NoMethodError.  Default to 1.
+        incident_id = 1
```
reidmix: I'd make this a test like:  `incident_id ||= 1 if Incident.synced.empty?` and do that before the method that'd throw.  Its more explicit that way and won't conceal other NoMethodErrors that could occur during a refactor etc.

lib/chat_ops/commands/start_incident_command.rb
```diff
@@ -1,3 +1,77 @@
...
+      ChatOps.help.each_line do |line|
+        message << "  " + line
+      end
+
+      { message: message.join("\n") }
+    end
+
+    private
+    def infer_incident_id(incident_id=nil)
+      # If they specified an incident ID, use it.
```
reidmix: You could also add `return incident_id if incident_id`

lib/chat_ops/commands/start_incident_command.rb
```diff
@@ -1,3 +1,77 @@
...
+        incident_id = 1
+      end
+
+      incident_id
+    end
+
+    def get_chat_start(timestamp=nil)
+      if !timestamp.blank?
+        ::IncidentResponse.parse_timestamp(timestamp)
+      end || Time.now
```
reidmix: How about an `else` block?

reidmix: You can also use `timestamp.present?` or flip it so 
```ruby
if timestamp.blank?
  Time.now
else
  ::IncidentResponse.parse_timestamp(timestamp)
end
```

lib/incident_response.rb
```diff
@@ -20,7 +20,58 @@ def prevent_highlights(message)
...
+          tz = fix_dst($1.upcase)
+
+          return unless result = Chronic.parse(timestamp)
+
+          result = Chronic.parse(result.strftime("%F %T #{tz}"))
+        end
+
+        return result
+      end
+    end
```
reidmix: Bleahg 😳

app/controllers/api/v1/chat_ops_controller.rb
```diff
@@ -7,7 +7,8 @@ def matcher
   end
 
   def respond
-    result = ChatOps.process(params['user'], params['message'])
+    user = User.ensure(**params.permit(user: [:email, :name, :handle])[:user].to_h.symbolize_keys)
```
lexelby: Yeah, I learned too :)

app/models/user.rb
```diff
@@ -10,4 +10,13 @@ def as_json(options={})
       handle: handle
     }
   end
+
+  class << self
+    def ensure(email:, name: nil, handle: nil)
+      user = find_or_create_by(email: email)
+      user.name = name
+      user.handle = handle
+      user.save!
```
lexelby: It's to ensure that we always have the most up-to-date name/handle if the user changes it in slack.

lib/chat_ops/commands/start_incident_command.rb
lexelby: Noted :)

lib/chat_ops/commands/start_incident_command.rb
lexelby: Hmm, interesting idea.  It also answers the question I've had running through my head, "Why are we instantiating the command class instead of just using class methods?"  Answer: because `run` sets variables used by `message`, saying what happened.

lib/chat_ops/commands/start_incident_command.rb
lexelby: thanks!

lib/chat_ops/commands/start_incident_command.rb
lexelby: Sure.  I just wanted to avoid the extra line.  **shrug**

lib/chat_ops/commands/start_incident_command.rb
lexelby: Hmm, but that kind of muddies _why_ we're setting it to 1.  I think I'll use an if/else instead.

lib/chat_ops/commands/start_incident_command.rb
lexelby: I can't use an `else` block.  I'm doing something kind of subtle here: I want to return Time.now if parsing fails, too.  In that case, IncidentResponse.parse_timestamp returns `nil`.

lexelby: I guess arguably it could return an exception on parse failure instead?

lib/incident_response.rb
lexelby: Bleahg what, the complciation?

app/models/user.rb
reidmix: Perfect.  

If that weren't the case there's the `where(email: name).first_or_create(name: name, handle: handle)` form, but that only updates name, handle on create.

lib/chat_ops/commands/start_incident_command.rb
reidmix: Does it?  Maybe a good comment will clarify.

lib/chat_ops/commands/start_incident_command.rb
reidmix: Oh I see, gotcha.  If it fails or there's no `timestamp`.

I think its fine in this case -- though it still looks wonky to get what's happening immediately -- which goes to show that I didn't pick up that subtlety.

lib/incident_response.rb
reidmix: Yeah, I'm not expecting to change anything, but I'd probably go and fix `Chronic` to do what you want than to suffer all these sidecases.

lexelby: Not really viable.  Chronic is unmaintained and there are already a ton of forks.  I'm not keen on the idea of maintaining our own fork and getting upstream to merge a PR seems like it's not gonna happen.

PR #45, opened 2016-09-09
cleaning up dot env stuff

jmervine:
Per [this comment](https://github.com/heroku/retrodot/pull/38#issuecomment-245590912), I'm doing a quick clean up of the `.env*` files.

1. Quotes are unnecessary.
1. `export` is unnecessary.
1. When using `.env.development`, we really don't need `.env`, thus I'm removing `.env.sample`.
    - A possible use-case for adding this back would be someone running on **not** heroku, where in they want/need to use `fore{man,go}`. However, I think it's safe to assume if someone is using `fore{man,go}`, they'll know enough to create their own `.env` file.
1. Updating `app.json` w/ variables referenced in sample files.

cc @heroku/sre


joshuatobin:
@jmervine this looks good but you need to update the `.travis.yml` as well.

reidmix:
Thanks for picking this up @jmervine /cc @lexelby 

You do have to update the `.travis.yml` to check on the existence of `.env.sample` before copying (FYI, that was the originating problem):
```
The command "cp .env.test{.sample,} && cp .env{.sample,}" failed and exited with 1 during .
```

reidmix:
👍

PR #46, opened 2016-09-09
Update retrodot with loggerator

joshuatobin:
CC: @reidmix @lexelby


lexelby:
:+1:

Should we remove the parentheses to be all ruby-style?

joshuatobin:
I personally have a preference for the parentheses but I guess lets just pick one way or the other and be consistent.

reidmix:
👍

PR #47, opened 2016-09-12
Sort activeadmin incidents by id

joshuatobin:
CC: @lexelby @reidmix


reidmix:
I think it looks good -- don't know ActiveAdmin that well.

lexelby:
Can't test it due to auth issues (will try to look into that today), but this looks good to me.  :+1:

PR #48, opened 2016-09-14
move IncidentResponse functions into ChatOps

lexelby:
I decided that `IncidentResponse` was a terrible name for chatops library functions.  I moved them to ChatOps.


joshuatobin:
👍

PR #43, opened 2016-09-06
Ir fixes

lexelby:
A couple little fixes and tests for bugs that appeared while demoing to the rest of the SRE team (of course).  Also adds logging to `rake sync:users`.

@jmervine @reidmix @joshuatobin

PR #49, opened 2016-09-14
Ir fixes

lexelby:
Redo of #43 because it ended up conflicting heavily with master.  Just a few misc fixes.


lexelby:
@tobin could you check this out when you have a chance?  Should be quick.

joshuatobin:
👍

app/support/mediators/base.rb
```diff
@@ -1,10 +1,15 @@
-include Loggerator
 
 module Mediators
   class Base
+    include Loggerator
 
     def self.run(options={})
-      new(options).call
+      log_context(mediator: self.name) do
```
reidmix: I said this on another PR but another way to do this (still not documented) was to `include Loggerator::Namespace` which will adds `ns: Mediators::Base` in the same way that `mediator: Mediators::Base` does here. 

Right now I'd have to amend it to work with class-level methods, which it currently doesn't do.

PR #50, opened 2016-09-15
End incident command

lexelby:
This implements `end incident`.  I added tests, and while I was in there I added a helper for writing tests for commands.  Then I went back and added a few more tests for `start incident`.  This PR kind of got away from me.

### DIFF
--- a/lib/chat_ops/chat_ops.rb
+++ b/lib/chat_ops/chat_ops.rb
@@ -90,6 +90,10 @@ module ChatOps
       end
     end
 
+    def current_incident
+      Incident.by_timeline_start&.first
+    end
+
     private
     # If passed "EST" during daylight time, return "EDT", and the reverse.
     # People often say "3pm EST" when it's technically "3pm EDT" during that
--- a/lib/chat_ops/commands/end_incident_command.rb
+++ b/lib/chat_ops/commands/end_incident_command.rb
@@ -7,7 +7,7 @@ module ChatOps::Commands
       incident = if match['incident_id']
         Incident.find_by(incident_id: match['incident_id'])
       else
-        IncidentResponse::current_incident()
+        ChatOps.current_incident()
       end
       return { message: 'unknown incident (do you need to "start incident" first?)' } unless incident
 
--- a/lib/chat_ops/commands/end_incident_command.rb
+++ b/lib/chat_ops/commands/end_incident_command.rb
@@ -1,5 +1,20 @@
 module ChatOps::Commands
   class EndIncidentCommand < ChatOps::ChatOpsCommand
-    match /end incident/
+    match /end\s+(the\s+)?incident(\s+(?<incident_id>[0-9]+))?/
+    help_message "end incident [#] - sets the end of chat for an incident"
+
+    def run(user, match)
+      incident = if match['incident_id']
+        Incident.find_by(incident_id: match['incident_id'])
+      else
+        IncidentResponse::current_incident()
+      end
+      return { message: 'unknown incident (do you need to "start incident" first?)' } unless incident
+
+      incident.chat_end = Time.nowq
+      incident.save
+
+      return { message: "Recorded the end of chat for incident #\#{incident.incident_id} at #{incident.chat_end.inspect}." }
+    end
   end
 end
--- a/lib/chat_ops/commands/end_incident_command.rb
+++ b/lib/chat_ops/commands/end_incident_command.rb
@@ -1,3 +1,5 @@
-class EndIncidentCommand < ChatOpsCommand
-  match /end incident/
+module ChatOps::Commands
+  class EndIncidentCommand < ChatOps::ChatOpsCommand
+    match /end incident/
+  end
 end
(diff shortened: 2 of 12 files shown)


lib/chat_ops/commands/end_incident_command.rb
```diff
@@ -1,3 +1,20 @@
-class EndIncidentCommand < ChatOpsCommand
-  match /end incident/
+module ChatOps::Commands
+  class EndIncidentCommand < ChatOps::ChatOpsCommand
+    match /end\s+(the\s+)?incident(\s+(?<incident_id>[0-9]+))?/
+    help_message "end incident [#] - sets the end of chat for an incident"
+
+    def run(user, match)
+      incident = if match['incident_id']
+        Incident.find_by(incident_id: match['incident_id'])
```
joshuatobin: Maybe its just me but the syntax for this reads strange.

spec/helpers/time_helper.rb
```diff
@@ -0,0 +1,5 @@
+RSpec::Matchers.define :match_to_the_millisecond do |expected|
+  match do |actual|
+    (actual - expected).abs < 0.001
+  end
```
joshuatobin: Oh this is neat. http://s2.quickmeme.com/img/ab/ab8c229496701a3c93896c03d0a251dc00d0428b5e74c8f1da2d40985648dca6.jpg

lexelby: :D

lib/chat_ops/commands/end_incident_command.rb
lexelby: because of using inline `if`?

joshuatobin: @lexelby nitpick - would you mind indenting the if statement so it lines up?

spec/lib/chat_ops/commands/start_incident_command_spec.rb
```diff
@@ -1,30 +1,19 @@
...
   end
 
   describe ".run" do
-    let(:start_incident_command) { ChatOps::Commands::StartIncidentCommand.new }
-    let(:user) { create(:user) }
-
-    def process(command)
-      start_incident_command.process(user, command)
-    end
+    command :start_incident
```
joshuatobin: nice

spec/lib/chat_ops/commands/end_incident_command_spec.rb
```diff
@@ -0,0 +1,39 @@
+RSpec.describe ChatOps::Commands::EndIncidentCommand do
+  include ChatOpsCommandHelper
+
+  describe "regex" do
+    test_regex_against_commands <<-EOL
+      end incident
+      end the incident
+      end incident 12
+    EOL
```
joshuatobin: This is pretty great.

joshuatobin:
👍

lib/chat_ops/commands/end_incident_command.rb
lexelby: I'm gonna skip it for now because I plan to DRY that into its own function in the next PR.

lib/chat_ops/chat_ops.rb
```diff
@@ -90,6 +90,10 @@ def parse_timestamp(timestamp)
       end
     end
 
+    def current_incident
+      Incident.by_timeline_start&.first
+    end
```
reidmix: The `by_timeline_start` will return an array even if it's empty.  You can just call `first` on it and be good.

FWIW this, too, could be a scope on incident:
```ruby
scope :current, { by_timeline_start.limit(1).first }
```

lib/chat_ops/commands/start_incident_command.rb
```diff
@@ -1,5 +1,5 @@
 module ChatOps::Commands
-  class StartIncidentCommand < ChatOpsCommand
+  class StartIncidentCommand < ChatOps::ChatOpsCommand
```
reidmix: A suggestion: might be simpler to now rename `ChatOpsCommand` to just `ChatOps::Command`

lib/chat_ops/chat_ops.rb
lexelby: Actually, that doesn't work.  Apparently `.first` on an empty result in a scope results in `[]` -- I guess a scope always returns a result set.  I could make it a class method but at that point it's just a matter of relocating the implementation of `ChatOps.current_incident`, and I don't think there's any benefit.

reidmix: I just checked locally, the code works without the lonely operator and the scope would work. 
If anything, I'd get rid of the lonely operator.  

@lexelby So here's my thought about not making a branch because (possibly) the change is too small to warrant it.  Maybe what could happen is one branch against all PRs and that would lower the bar to make changes?

lexelby: Great idea.  Unfortunately I missed this one in the PR I did for a bunch of your suggestions :(

reidmix: It's a "Next time you see it" kinda-thing :D

PR #51, opened 2016-09-15
Add support for trello card creation

joshuatobin:
Adds a mediator for creating trello cards via the Admin UI. 

CC: @lexelby @reidmix


app/admin/incident.rb
```diff
@@ -1,3 +1,4 @@
+require 'trello'
```
lexelby: Is this require necessary?  Shouldn't the gemfile take care of that?

app/admin/incident.rb
```diff
@@ -11,10 +12,29 @@
 
   # This creates /admin/:incident/sync which is linked to below.
   member_action :sync, method: :post do
-    Mediators::Incident::Syncher.run(incident: resource[:incident_id])
+    Mediators::Incident::OneSyncher.run(id: resource[:incident_id])
```
lexelby: Oops, thanks!

app/controllers/trello_controller.rb
```diff
@@ -0,0 +1,18 @@
+class TrelloController < ApplicationController
+  skip_before_action :verify_authenticity_token, only: :callback
```
lexelby: There is no `callback` function.  `only: :create`, perhaps?  I sort of wonder why this works even though this is wrong :)

app/support/mediators/incident/create_card.rb
```diff
@@ -0,0 +1,31 @@
...
+        oauth_token_secret: @oauth_secret
+      )
+
+      @template ||= @trello.find(:card, Config.trello_template)
+
+      @trello.create(:card,
+        "name"         => "Incident #{@id}: #{@title}",
+        "idCardSource" => @template.id,
+        "idList"       => @template.list_id
+      )
```
lexelby: The existing code not only copies the template, it also does some text search and replace, starting at line 359.

config/initializers/omniauth.rb
```diff
@@ -1,4 +1,5 @@
 Rails.application.config.middleware.use OmniAuth::Builder do
   provider :developer if Rails.env.development? || Config.pr_app?
   provider :google_oauth2, Config.google_client_id, Config.google_client_secret, { hd: Config.google_domain }
+  provider :trello, Config.trello_consumer_key, Config.trello_consumer_secret, app_name: Config.app_name, scope: "read,write,account", expiration: "never"
```
lexelby: I'm pretty sure this is okay, but check to make sure that if a logged out user goes directly to `/auth/trello` they don't end up logged into retrodot.

app/admin/incident.rb
joshuatobin: @lexelby Yeah things blow up if that wasn't required

app/controllers/trello_controller.rb
joshuatobin: whoops. originally called it callback, but then renamed

joshuatobin: Fixed in 975521a

app/support/mediators/incident/create_card.rb
joshuatobin: @lexelby Yeah we don't have any of that information yet. Once the Google apps logic get sorted, we'll need to come back and fix.

lexelby: oh right, heh

app/admin/incident.rb
lexelby: Weird.

reidmix: I just rechecked this and it works now without requiring (which was my expectation as well). @joshuatobin

app/controllers/trello_controller.rb
```diff
@@ -0,0 +1,20 @@
+class TrelloController < ApplicationController
+  skip_before_action :verify_authenticity_token, only: :create
+
+  def create
+    return redirect_to "/auth/failure" unless current_user
+
+    session['trello-oauth-token']  = auth_hash["credentials"]["token"]
+    session['trello-oauth-secret'] = auth_hash["credentials"]["secret"]
+
+    log(fn: "create", at: "update_oauth", user: session[:user][:email])
```
reidmix: Can/should we add something here in the log to indicate its the trello auth?

app/controllers/trello_controller.rb
```diff
@@ -0,0 +1,20 @@
...
+    current_user.update(trello_oauth_token: session['trello-oauth-token'], trello_oauth_secret: session['trello-oauth-secret'])
```
reidmix: Does the `trello_oauth_*` need to be both in the session and in the database?

app/support/mediators/incident/create_card.rb
```diff
@@ -0,0 +1,31 @@
+require 'trello'
+module Mediators::Incident
+  class CreateCard < Mediators::Base
```
reidmix: Just as an FYI, you can add `include Loggerator::Namespace` in `Mediators::Base` and it will add "ns: Mediators::Incident::Incident" to the logs automatically.   

I see that `Mediators::Base` is doing this as well by adding a `log_context(mediator: self.name)` and that is perfectly good! Esp. as there was no documentation prior to doing this :D

app/support/mediators/incident/create_card.rb
reidmix: Create a card to do this if you haven't already <3

app/support/mediators/incident/create_card.rb
```diff
@@ -0,0 +1,31 @@
...
+      )
+
+      @template ||= @trello.find(:card, Config.trello_template)
+
+      @trello.create(:card,
+        "name"         => "Incident #{@id}: #{@title}",
+        "idCardSource" => @template.id,
+        "idList"       => @template.list_id
+      )
+    rescue Trello::InvalidAccessToken, Trello::Error, NoMethodError
```
reidmix: Curious as to rescuing `NoMethodError` here, as in what is the case this could happen.

I've found that rescuing `NoMethodError` can really disguise problems esp. in development (I see its getting logged and re-raised) but maybe its something that you'd be able to remove?

db/migrate/20160913152840_add_token_secret_to_users.rb
```diff
@@ -0,0 +1,6 @@
+class AddTokenSecretToUsers < ActiveRecord::Migration[5.0]
```
reidmix: woah, what is the `[5.0]` in `ActiveRecord::Migration[5.0]` I've never seen that.

reidmix: Ah figured it out ActiveRecord::Migration[5.0] (cray! and good thinking on rails' part)

spec/spec_helper.rb
```diff
@@ -7,6 +7,7 @@
 require File.expand_path("../../config/environment", __FILE__)
 require 'rspec/rails'
 require 'loggerator/test'
+require 'webmock/rspec'
```
reidmix: Curious, why did you have to move this out of the Gemfile?

joshuatobin: It was causing havoc and blowing up.

reidmix: weird, it should be required by the Gemfile -- can you explain the havoc?

joshuatobin: I can't remember the details TBH, it was last week. I know for sure it didn't work which is why I futzed with it.

reidmix: Thanks, I'll tinker and see wtf.

reidmix: Ah, it works with `rake` where it gets loaded with the environment, but `rspec` it does not for some reason.

reidmix: The actual reason this doesn't work is that webmock doesn't specify rspec as a dependency because `rspec` isn't required out of the box for webmock.  That's why you have to `require "webmock/rspec"`.   So when bundler figures out dependencies, `webmock` is loaded before `rspec/core` and it fails. 

The way you have it here is correct! 👍

PR #52, opened 2016-09-15
'Add person to incident' command

lexelby:
@joshuatobin here's another one -- sorry for all the review requests!


lib/chat_ops/chat_ops.rb
```diff
@@ -94,6 +94,32 @@ def current_incident
...
+
+    def error(text)
+      # In the future, this might tag with some kind of metadata indicating an
+      # error.
+      { message: text }
+    end
+
+    def message(text)
+      { message: text }
+    end
```
joshuatobin: Couldn't these just be one method for now?

lib/chat_ops/commands/end_incident_command.rb
```diff
@@ -4,12 +4,7 @@ class EndIncidentCommand < ChatOps::ChatOpsCommand
     help_message "end incident [#] - sets the end of chat for an incident"
 
     def run(user, match)
-      incident = if match['incident_id']
-        Incident.find_by(incident_id: match['incident_id'])
-      else
-        ChatOps.current_incident()
-      end
-      return { message: 'unknown incident (do you need to "start incident" first?)' } unless incident
+      incident = ChatOps.determine_incident(match['incident_id']) or return ChatOps.unknown_incident
```
joshuatobin: Did you have to use `or` here vs `||` because of the `return`?

lib/chat_ops/chat_ops.rb
lexelby: I suppose so. :)  I'll make it an alias but keep the comment.  That way we can turn it into a method that does something different if we want.

lib/chat_ops/commands/add_responder_command.rb
```diff
@@ -0,0 +1,17 @@
...
+
+    def run(user, match)
+      incident = ChatOps.determine_incident(match['incident_id']) or return ChatOps.unknown_incident
+
+      responders = ChatOps.get_mentioned_users(match['who'])
+      return ChatOps.error("Person not found.  Try using their @handle.") if responders.length == 0
+
+      incident.responders += responders
+
+      ChatOps.reaction(':checkmark:')
```
joshuatobin: <3

lib/chat_ops/commands/end_incident_command.rb
lexelby: I use `or` because I like the way it looks; no other reason.  I feel like this reads more like a sentence.  ¯\_(ツ)_/¯

spec/lib/chat_ops/commands/add_responder_command_spec.rb
```diff
@@ -0,0 +1,34 @@
+RSpec.describe ChatOps::Commands::AddResponderCommand do
+  include ChatOpsCommandHelper
+
+  describe 'regex' do
+    test_regex_against_commands <<-EOL
+      add person to incident
+      add person and other person to incident
+      add person to incident 12
+      add person and other person to incident 12
```
joshuatobin: what about `add person1 person2 person3 to incident`

lib/chat_ops/commands/end_incident_command.rb
joshuatobin: Actually using `or` here is probably the right way since you're using it as flow control. I tend to shy away from using it though due to precedence. http://www.virtuouscode.com/2010/08/02/using-and-and-or-in-ruby/

lib/chat_ops/chat_ops.rb
joshuatobin: 👍

joshuatobin:
👍

lib/chat_ops/chat_ops.rb
```diff
@@ -94,6 +94,30 @@ def current_incident
...
+      message 'unknown incident (do you need to "start incident" first?)'
+    end
+
+    def message(text)
+      { message: text }
+    end
+
+    # In the future, this might tag with some kind of metadata indicating an
+    # error.
+    alias_method :error, :message
```
joshuatobin: 👍

joshuatobin:
👍

lib/chat_ops/commands/end_incident_command.rb
lexelby: Yup.  I tested the precedence first just to be sure.

spec/lib/chat_ops/commands/add_responder_command_spec.rb
lexelby: Well, I have a test for two people in the list.  I don't really think I need to test all the permutations since it uses `ChatOps.get_mentioned_users` which I tested the hell out of.

lib/chat_ops/chat_ops.rb
```diff
@@ -94,6 +94,30 @@ def current_incident
...
+
+    def reaction(name)
+      { reaction: name }
+    end
+
     private
     # If passed "EST" during daylight time, return "EDT", and the reverse.
     # People often say "3pm EST" when it's technically "3pm EDT" during that
```
reidmix: I'm feeling like this is a utility class that should be a module that's mixed into `ChatOps::ChatOpsCommand`
I find that utility classes end up being a dumping ground for code that one can't figure out a good place for (on other projects, such as the `Utils` class in Status). 

Seems like these methods in this module are used on such a regular basis in `ChatOps::ChatOpsCommand` subclasses that maybe it should just be available to them?

You can see this below in the `AddResponderCommand` class where `ChatOps` used on (nearly) every line in the `run` method.

lib/chat_ops/commands/end_incident_command.rb
reidmix: FWIW, most people have opted out of using `and`/`or` -- I also like how it looks and how the precedence works, but mostly been on projects where they are unfavorable.

lexelby: Meh, call it a pythonism.  I'm keeping it. ;) :rebel:

lib/chat_ops/chat_ops.rb
lexelby: Hmm, good point.  Would it make sense to just move the methods to ChatOps::Command (formerly ChatOpsCommand)?

reidmix: Yeah, that would work too 👍 If you end up having many commands, an organizational pattern is to group them by function into several modules and include them together -- but that's when files become really unwieldy.

lib/chat_ops/commands/end_incident_command.rb
reidmix: HA! Go for it.

PR #53, opened 2016-09-15
List incident responders

joshuatobin:
CC: @lexelby @reidmix


lib/chat_ops/commands/list_responder_command.rb
```diff
@@ -0,0 +1,12 @@
+module ChatOps::Commands
+  class ListResponderCommand < ChatOps::ChatOpsCommand
+    match /list\s+incident(\s+(?<incident_id>\d+))?\s+responders/
+    help_message "h incident [#] responders - list responders for an incident"
+
+    def run(user, match)
+      incident = ChatOps.determine_incident(match['incident_id']) or return ChatOps.current_incident
+
+      ChatOps.message(incident.responders.map(&:handle).join(" "))
```
lexelby: Can we give feedback about what we're talking about?  How about "Incident responders for incident #n: <list>".  That also has the advantage of not sending an empty message in the (unlikely) event that the responders list is empty.

lib/chat_ops/commands/list_responder_command.rb
```diff
@@ -0,0 +1,12 @@
+module ChatOps::Commands
+  class ListResponderCommand < ChatOps::ChatOpsCommand
+    match /list\s+incident(\s+(?<incident_id>\d+))?\s+responders/
+    help_message "h incident [#] responders - list responders for an incident"
+
+    def run(user, match)
+      incident = ChatOps.determine_incident(match['incident_id']) or return ChatOps.current_incident
```
lexelby: Bug here: it should return `ChatOps.unknown_incident` on the right side rather than `current_incident`.  If you copied this code from somewhere then we should fix that place too...

lib/chat_ops/commands/list_responder_command.rb
```diff
@@ -0,0 +1,12 @@
+module ChatOps::Commands
+  class ListResponderCommand < ChatOps::ChatOpsCommand
+    match /list\s+incident(\s+(?<incident_id>\d+))?\s+responders/
```
lexelby: In the current bot, the command is just: `h incident responders`, so "list" should be removed here.

lib/chat_ops/commands/list_responder_command.rb
lexelby: Also, the current code uses real names (comma separated) instead of handles.

lexelby: Oh!! And run the message through `ChatOps.prevent_highlights` too just to make sure we don't ping everyone.  Perhaps `ChatOps.message` should do this automatically.

spec/lib/chat_ops/commands/list_responder_command_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+
+      expect(process("list incident responders")).to return_response_matching /#{user1.handle}/
+    end
+
+    it "should list multiple responders" do
+      incident.responders << user1
+      incident.responders << user2
+      incident.save
+
+      expect(process("list incident responders")).to return_response_matching /#{user1.handle} #{user2.handle}/
```
lexelby: This test is kind of brittle because it depends on `Incident.responders` returning users in the order added.  Maybe `.to return_response_matching(/#{user1.handle}/).and return_response_matching(/#{user2.handle}/)`?

lib/chat_ops/commands/list_responder_command.rb
joshuatobin: +1 automagically

joshuatobin:
@lexelby pushed suggested changes

lib/chat_ops/commands/list_responder_command.rb
joshuatobin: Fixing this in another PR btw

spec/lib/chat_ops/commands/list_responder_command_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+
+      expect(process("incident responders")).to return_response_matching /#{user1.handle}/
+    end
+
+    it "should list multiple responders" do
+      incident.responders << user1
+      incident.responders << user2
+      incident.save
+
+      expect(process("incident responders")).to return_response_matching(/#{user1.handle}/).and return_response_matching(/#{user2.handle}/).and return_response_matching /#{incident.id}/
```
lexelby: just out of curiosity... can you try it like this?

`.to return_response_matching(/#{user1.handle}/, /#{user2.handle}/, /#{incident.incident_id}/)`

I think that might be automatically supported by rspec.  Also note that it should be `incident.incident_id`.

lib/chat_ops/commands/list_responder_command.rb
```diff
@@ -0,0 +1,12 @@
+module ChatOps::Commands
+  class ListResponderCommand < ChatOps::ChatOpsCommand
+    match /incident(\s+(?<incident_id>\d+))?\s+responders/
+    help_message "h incident [#] responders - list responders for an incident"
+
+    def run(user, match)
+      incident = ChatOps.determine_incident(match['incident_id']) or return ChatOps.unknown_incident
+
+      ChatOps.message("Incident responders for incident #{incident.incident_id}: #{incident.responders.map(&:name).join(", ")}")
```
reidmix: suggestion: make a `responder_names` method in `Incident`

PR #56, opened 2016-09-16
DRY incident ID parsing in commands

lexelby:
Move parsing of incident ID to ChatOpsCommand.  If a command's `run` method takes a parameter `incident`, then ChatOpsCommand will parse the incident id (or default to the current incident) and pass it into the command.  This is a big win for DRYing incident commands.

cc @joshuatobin

### DIFF
--- a/lib/chat_ops/chat_ops_command.rb
+++ b/lib/chat_ops/chat_ops_command.rb
@@ -21,7 +21,12 @@ module ChatOps
 
     def process(user, message)
       if result = self.class.regex.match(message)
-        run(user, result)
+        if should_parse_incident
+          incident = ChatOps.determine_incident(result[:incident_id]) or return ChatOps.unknown_incident
+          run(user, result, incident)
+        else
+          run(user, result)
+        end
       end
     end
 
@@ -36,9 +41,26 @@ module ChatOps
     #
     # Return nil to indicate that we don't actually want to process the command
     # after all.
+    #
+    # If the run method takes an `incident` parameter, then it is a command that
+    # allows the user to optionally pass an incident ID.  The regex should have
+    # a named capture group like (?<incident_id>\d+)?.  If the user passed an
+    # incident ID, it will be looked up and the incident object will be passed
+    # into run().  Otherwise the current incident is looked up and passed in
+    # (see ChatOps.current_incident).
 
     def run(user, match_data)
       raise NotImplementedError
+    end
+
+    private
+    # If the run method takes a parameter named `incident`, then this is a
+    # command that optionally allows the incident to be passed in.
+    def should_parse_incident
+      run = method :run
+      run.parameters.any? do |type, name|
+        name == :incident
+      end
     end
   end
 end
--- a/lib/chat_ops/chat_ops_command.rb
+++ b/lib/chat_ops/chat_ops_command.rb
@@ -1,7 +1,7 @@
 module ChatOps
   class ChatOpsCommand
     class << self
-      attr_reader :regex, :name, :help
+      attr_reader :regex, :name, :help, :should_parse_incident
 
       # Ruby calls this function when a class is declared that inherits from this
       # class.  We then register it with the ChatOps module.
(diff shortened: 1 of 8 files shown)


lexelby:
I'll wait until the existing command PRs land and merge them in here and DRY them before merging this.

lib/chat_ops/chat_ops_command.rb
```diff
@@ -36,9 +41,26 @@ def process(user, message)
...
       nil
     end
+
+    private
+    # If the run method takes a parameter named `incident`, then this is a
+    # command that optionally allows the incident to be passed in.
+    def should_parse_incident
+      run = method :run
+      run.parameters.any? do |type, name|
+        name == :incident
```
joshuatobin: Should we add another check here to make sure name name isn't nil `!!name.nil? && name == :incident` so we can pass in incident: nil in the generator?

lib/chat_ops/chat_ops_command.rb
```diff
@@ -17,11 +17,25 @@ def match(r)
...
+      # ChatOps.current_incident is used.
+      def parse_incident(should_parse_incident)
+        @should_parse_incident = should_parse_incident
+      end
     end
 
     def process(user, message)
       if result = self.class.regex.match(message)
-        run(user, result)
+        if self.class.should_parse_incident
```
joshuatobin: so much better. thanks

lib/chat_ops/commands/end_incident_command.rb
```diff
@@ -1,11 +1,10 @@
 module ChatOps::Commands
   class EndIncidentCommand < ChatOps::ChatOpsCommand
     match /end\s+(the\s+)?incident(\s+(?<incident_id>[0-9]+))?/
+    parse_incident true
     help_message "end incident [#] - sets the end of chat for an incident"
```
joshuatobin: @lexelby mind just adding this to the generator as a comment?

lexelby:
@tobin rebased on master and added `parse_incident` to the generator.

joshuatobin:
🚢

lib/chat_ops/chat_ops_command.rb
reidmix: What if you try to parse the incident every time and just make it available if it is, or nil if it isn't?
That way you have the same process behavior for all commands and you don't have to worry about this ` if self.class.should_parse_incident' check

lib/chat_ops/commands/add_responder_command.rb
```diff
@@ -1,11 +1,10 @@
 module ChatOps::Commands
   class AddResponderCommand < ChatOps::ChatOpsCommand
     match /add\s+(?<who>.*)\s+to\s+incident(\s+(?<incident_id>\d+))?/
+    parse_incident true
     help_message "h add <person> [<person>...] to incident -- <person> can be a full name or @handle"
 
-    def run(user, match)
-      incident = ChatOps.determine_incident(match['incident_id']) or return ChatOps.unknown_incident
-
+    def run(user, match, incident)
```
reidmix: Nice change for sure 👍

lib/chat_ops/chat_ops_command.rb
lexelby: I thought of that.  The problem is `h start incident`.  It has an `incident_id` but it's okay if that incident doesn't exist.  However for all the other commands, if an `incident_id` is specified by the user and that incident doesn't exist, we want to return `ChatOps.unknown_incident`.  If we get rid of that then we lose some DRYness.

reidmix: Ah, I got it.  So, just return `Chatops.unknown_incident` as you are doing in process but only when  `should_parse_incident` -- so it's just re-organizing a little bit to get the unhappy-paths out of the way up front:

```ruby
def process(user, message)
  return unless result = self.class.regex.match(message)
  
  incident = ChatOps.determine_incident(result[:incident_id]) 
  return ChatOps.unknown_incident if self.class.should_parse_incident && !incident
  
  run(user, result, incident)
end  
```

lexelby: I see the value here, but I think at this point I'd rather stop messing around with this code and move on.

lexelby: I decided to try to implement what you've suggested here, but I got mired in special cases that made the whole thing less readable in the end.  We want most commands to just get an Incident object for the incident the user specified.  But in the case of `start incident 123`, if incident 123 doesn't exist, what should we pass in? Just the integer 123?  But it's also legit to do `h start incident 123` if incident 123 was already opened to overwrite the start time.  Do we pass in the Incident object if it exists, otherwise the integer?  Messy.

reidmix: In the case of `StartIncidentCommand`, I'd expect that `StartIncidentCommand` will still use `infer_incident_id(match[:incident_id])` in this case to get the `incident_id` and create the incident. 

I don't even think it matters what `determine_incident` does in this case (so incident could be `nil`)

PR #58, opened 2016-09-16
Prevent highlights in chatops messages 

joshuatobin:
* Prevent highlights in chatops messages 
* Return NotImplemented for run


lib/chat_ops/chat_ops_command.rb
```diff
@@ -38,7 +38,7 @@ def process(user, message)
     # after all.
 
     def run(user, match_data)
-      nil
+      raise NotImplementedError
```
reidmix: Nice!

PR #62, opened 2016-09-26
Collect google auth tokens on login

joshuatobin:
This PR is a pre-requisite for adding support for google drive integration. This will allow us to collect auth and refresh tokens on login and update them if needed. 

CC: @lexelby @reidmix

### DIFF
--- a/app/controllers/auth_controller.rb
+++ b/app/controllers/auth_controller.rb
@@ -6,10 +6,9 @@ class AuthController < ApplicationController
   def callback
     log(fn: 'callback')
     unless session['user'].present?
-      user = request.env['omniauth.auth']['info']
-
+      user  = request.env['omniauth.auth']['info']
       email = (user['email']||'').downcase
-      name = user['name']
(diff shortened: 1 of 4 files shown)


app/controllers/auth_controller.rb
```diff
@@ -47,14 +45,25 @@ def unauthorized
...
+    render plain: 'You are not authorized', status: 401
   end
 
   def verify
     head(:ok)
   end
 
+  def developer?
+    request.env['omniauth.auth']['provider'] == "developer"
+  end
```
reidmix: Nice!

app/controllers/auth_controller.rb
```diff
@@ -47,14 +45,25 @@ def unauthorized
   end
 
   def failure
-    render text: 'You are not authorized', status: 401
+    render plain: 'You are not authorized', status: 401
```
reidmix: That depreciation warning, so interesting.

PR #64, opened 2016-09-27
Mediator for google script execution 

joshuatobin:
CC: @lexelby @reidmix


reidmix:
Looks good to me except for needing tests.

PR #65, opened 2016-09-30
Add mediators for preparing a retro

joshuatobin:
This PR adds a few new mediators for preparing for a retro by introducing support for the google api execution service. Still working on tests but opening early since @lexelby is out next week.

CC: @reidmix


app/admin/incident.rb
```diff
@@ -16,26 +18,29 @@
...
-  # /admin/:incident/create_trello_card
-  member_action :create_trello_card, method: %i[get post] do
+  member_action :prepare_retro, method: %i[get post] do
     begin
-      Mediators::Incident::CreateCard.run(
-        id:                  resource[:incident_id],
-        title:               resource[:title],
-        trello_oauth_token:  current_user.trello_oauth_token,
-        trello_oauth_secret: current_user.trello_oauth_secret
+      incident = Incident.find_by(id: resource[:id])
```
lexelby: `resource` here should actually be the Incident.  You could just pass it in instead.

app/admin/incident.rb
```diff
@@ -16,26 +18,29 @@
...
+      Mediators::Incident::PrepareRetro.run(
+        incident:     incident,
+        current_user: current_user
       )
     rescue Trello::InvalidAccessToken, Trello::Error, NoMethodError
       log_error($!, at: :create_trello_card, fn: :member_action)
-      session[:return_to] = resource[:incident_id]
+      session[:return_to] = resource[:id]
       redirect_to "/auth/trello"
+    rescue Signet::AuthorizationError, Google::Apis::AuthorizationError
```
lexelby: I'd rather this code didn't have to know the specific Trello and Google exceptions that are thrown by their respective gems.  The mediator is where the logic about deciding what to do with those gems' exceptions should be.

You could have the mediator declare two exception classes, `TrelloAuthRequired` and `GoogleAuthRequired`, and just catch those here.

app/admin/incident.rb
```diff
@@ -45,8 +50,8 @@
     link_to 'Sync', sync_admin_incident_path(resource)
   end
 
-  action_item :create_trello_card, only: :post do
-    link_to 'Create Trello Card', create_trello_card_admin_incident_path(resource)
+  action_item :prepare_retro, only: :post do
+    link_to 'Prepare Retro Card/Doc', prepare_retro_admin_incident_path(resource)
```
lexelby: Is there a reason this one is a GET whereas the one below is a POST?  IMO they should both be POST.

app/controllers/trello_controller.rb
```diff
@@ -8,9 +8,12 @@ def create
...
 
     log(fn: "create", at: "update_oauth", user: session[:user][:email])
-    current_user.update(trello_oauth_token: session['trello-oauth-token'], trello_oauth_secret: session['trello-oauth-secret'])
+    current_user.update(
+      trello_oauth_token:  session['trello-oauth-token'],
+      trello_oauth_secret: session['trello-oauth-secret']
+    )
 
-    redirect_to create_trello_card_admin_incident_path(session.delete(:return_to))
+    redirect_to prepare_retro_admin_incident_path(session.delete(:return_to)
```
lexelby: missing close paren

app/support/mediators/incident/create_card.rb
```diff
@@ -18,11 +18,14 @@ def call
...
-        "idCardSource" => @template.id,
-        "idList"       => @template.list_id
+      @card = @trello.create(:card,
+        "name"           => "Incident #{@id}: #{@title}",
+        "idCardSource"   => @template.id,
+        "idList"         => @template.list_id,
+        "keepFromSource" => "all"
       )
+
+      @card
```
lexelby: Why are you saving this as an instance variable now?

app/support/mediators/incident/edit_card.rb
```diff
@@ -0,0 +1,34 @@
...
+      @card.desc = update_description
+      @card.update!
+
+      @card
+    rescue Trello::InvalidAccessToken, Trello::Error, NoMethodError
+      log_error($!, fn: "call", at: "run", card_id: @card_id)
+      raise
+    end
+
+    def update_description
```
lexelby: The old one does a search and replace using the existing card description, rather than replacing it entirely.

app/support/mediators/incident/prepare_retro.rb
```diff
@@ -0,0 +1,49 @@
+module Mediators::Incident
+  class PrepareRetro < Mediators::Base
+
+    def initialize(incident:, current_user:)
+      @incident, @current_user = incident, current_user
+    end
+
+    def call
```
lexelby: I like how clean this method became, just calling out to other mediators.  Nice work.

config/config.rb
```diff
@@ -40,4 +40,8 @@ module Config
   optional :trello_consumer_key, string
   optional :trello_consumer_secret, string
   optional :trello_template, string
+
+  # Google Script
+  optional :google_script_id, string
```
lexelby: Could you add a comment indicating what this script is for, and perhaps also check the script itself into the repo somewhere?

app/controllers/trello_controller.rb
joshuatobin: Missing a `)` here.

app/support/mediators/incident/create_card.rb
```diff
@@ -18,11 +18,14 @@ def call
...
 
-      @trello.create(:card,
-        "name"         => "Incident #{@id}: #{@title}",
-        "idCardSource" => @template.id,
-        "idList"       => @template.list_id
+      @card = @trello.create(:card,
+        "name"           => "Incident #{@id}: #{@title}",
+        "idCardSource"   => @template.id,
+        "idList"         => @template.list_id,
+        "keepFromSource" => "all"
```
joshuatobin: Even though the docs and code say this is the default, I still had to add it to get the checklists to copy over.

app/support/mediators/incident/create_retro_doc.rb
```diff
@@ -0,0 +1,27 @@
+require 'google/apis/script_v1'
+
+module Mediators::Incident
+  class CreateRetroDoc < Mediators::Base
+
+    def initialize(auth:, id:, title:, trello_url:, postmortem_date: false, dev_mode: false)
```
joshuatobin: I'd like to just have this take `**args` but for another time.

app/support/mediators/incident/edit_card.rb
```diff
@@ -0,0 +1,34 @@
...
+      "[Heroku status incident](https://status.heroku.com/admin/incidents/#{@id})\n\n" \
+      "[Gdoc notes](#{@gdoc})\n\n"
```
joshuatobin: Hard coding this but its going to get ripped out eventually as the trello integration is somewhat useless.

app/support/mediators/incident/create_card.rb
joshuatobin: This needs to return `id` and `url`

app/support/mediators/incident/edit_card.rb
joshuatobin: /shrug. This was easier and is cleaner IMO.

app/support/mediators/incident/prepare_retro.rb
joshuatobin: 👍

config/config.rb
joshuatobin: I don't think we want to check the script into a public repo.

app/admin/incident.rb
joshuatobin: Not sure what you mean here? I had to do the `find_by` to get it to work. Maybe there is a better way but I couldn't seem to just pass in `resource` if thats what you mean?

app/admin/incident.rb
joshuatobin: Fixed.

app/support/mediators/incident/prepare_retro.rb
```diff
@@ -0,0 +1,52 @@
+class GoogleAuthRequired < StandardError; end
+class TrelloAuthRequired < StandardError; end
+
```
joshuatobin: @lexelby New exception classes

app/support/mediators/incident/edit_card.rb
lexelby: True.

app/support/mediators/incident/edit_card.rb
lexelby: ...but we're probably ripping out trello anyway.  Ignore me.

app/support/mediators/incident/create_card.rb
lexelby: Right, but why not just have `@trello.create(...` be the last statement so the card is returned directly?  Why save it to an instance variable and then return the variable?

config/config.rb
lexelby: Well, we already have the retrospective template itself posted publicly...

app/admin/incident.rb
joshuatobin: Nevermind. Im not sure why it wasn't working originally. Fixed.

joshuatobin:
@lexelby I believe I've addressed all your concerns. 

@reidmix can you take a look and let me know what you'd like to see in specs to get this merged down?

app/admin/incident.rb
reidmix: FWIW, another reason they should favor PUT/POST is because they are actions that create a change in the system.  For example, browsers that have prefetch functionality could accidentally create the trello/gdoc if they were to traverse the link.

app/support/mediators/incident/create_retro_doc.rb
```diff
@@ -0,0 +1,30 @@
...
+end
+
+module Mediators::Incident;
+  class CreateRetroDoc < Mediators::Base
+
+    def initialize(auth:, id:, title:, trello_url:, postmortem_date: false, dev_mode: false)
+      @auth, @id, @title, @trello_url, @postmortem_date, @dev_mode = auth, id, title, trello_url, postmortem_date, dev_mode
+
+      @service = Google::Apis::ScriptV1::ScriptService.new
+      @service.authorization = @auth
```
reidmix: you can keep `auth` a local variable here since it's not used outside of the method
ie. `auth` instead of `@auth`

app/support/mediators/incident/prepare_retro.rb
reidmix: These are also defined in `app/admin/incident.rb` -- you will likely get "constant already defined" warnings. 
And it will be better to have them in one place.

app/support/mediators/incident/prepare_retro.rb
reidmix: agreed, it's super clean.

lexelby:
Yup, my concerns look good now.  Just need to fix what reid pointed out :)

app/support/mediators/incident/create_retro_doc.rb
joshuatobin: 👍 fixed

lexelby:
:+1:

PR #66, opened 2016-09-30
suggest `start incident` in certain circumstances

lexelby:
We've had problems with folks forgetting to do `start incident` and just doing `timeline <message>`.  That results in the message being attached to the previous incident, which might've been from days ago.  This adds some heuristic checks to see if perhaps they didn't mean the last incident.  The user can avoid these checks by specifying the incident explicitly.

There's some gnarly metaprogramming in `setup_command` that was necessary to DRY the tests.


lexelby:
After lunch, I'll rebase and add a check for overriding and adding timeline entries to old incidents.

lib/chat_ops/command.rb
```diff
@@ -53,5 +54,29 @@ def process(user, message)
...
+      # the incident is still open in the external incident source, so they
+      # probably mean this one
+      return false if incident.open?
+
+      return true if !incident.timeline_entries.empty? && incident.timeline_entries.last.timestamp < 1.hour.ago
+
+      return true if incident.timeline_entries.empty? && incident.chat_start? && incident.chat_start < 1.hour.ago
+
+      false
+    end
```
reidmix: This could/should be a method on` Incident`.

lib/chat_ops/command.rb
```diff
@@ -15,7 +15,7 @@ def match(r)
       end
 
       def help_message(text)
-        @help = text
+        @help = Config.chatops_prefix + text
```
reidmix: Is there a separator (or will one needed to be added to `chatops_config`)?  Seems so, looking further down. If you keep the default as `nil`, doing this might remove the need to include a space in the `chatops_prefix` itself:
```ruby
[Config.chatops_prefix, text].compact.join(" ")
```

lib/chat_ops/command.rb
```diff
@@ -53,5 +54,29 @@ def process(user, message)
...
+      return false if incident.open?
+
+      return true if !incident.timeline_entries.empty? && incident.timeline_entries.last.timestamp < 1.hour.ago
+
+      return true if incident.timeline_entries.empty? && incident.chat_start? && incident.chat_start < 1.hour.ago
+
+      false
+    end
+
+    def old_incident(incident)
```
reidmix: Naming suggestion: `old_incident_warning`

lib/chat_ops/command.rb
lexelby: Oh, nifty idea.

PR #67, opened 2016-09-30
Chatops methods to command

lexelby:
Note that this depends on #66.

Lots of changes in here, but it boils down to this:

1. Move helper methods from `ChatOps` to `ChatOps::Command`.  (Actually, grouped related methods into modules and included them in `ChatOps::Command`.)
2. Instead of passing `user` and `match` to `Command.process` and then passing them in turn to `run()`, store them as instance variables.
3. Use the new instance variable semantics to change from `parse_incident` to always parsing, with `incident_optional` for exceptional commands like `start incident`.

All of this makes the code much better organized and easier to work with.

The rest of this PR is just a bunch of little changes to make the tests pass.

@joshuatobin @reidmix


joshuatobin:
I'm gonna hold off reviewing this until #66 is merged.

lib/chat_ops/command.rb
```diff
@@ -1,14 +1,24 @@
 module ChatOps
   class Command
     class << self
-      attr_reader :regex, :name, :help, :should_parse_incident
+      attr_reader :regex, :help
```
reidmix: perhaps make `match` and `incident` readers too.

lib/chat_ops/commands/remove_responder_command.rb
```diff
@@ -1,16 +1,15 @@
...
   class RemoveResponderCommand < ChatOps::Command
     match /remove\s+(?<who>.*)\s+from\s+incident(\s+(?<incident_id>\d+))?/
-    parse_incident true
     help_message "remove <person> [<person>...] from incident -- <person> can be a full name or @handle"
 
-    def run(user, match, incident)
-      responders = ChatOps.get_mentioned_users(match['who'])
-      return ChatOps.error("Person not found.  Try using their @handle.") if responders.length == 0
+    def run
+      responders = get_mentioned_users(@match['who'])
```
reidmix: Since these always seem to come from the `match`, you could just pass in the name `:who or "who"`

lib/chat_ops/command.rb
```diff
@@ -18,25 +28,34 @@ def help_message(text)
...
+
+          if !self.class.incident_optional?
+            return unknown_incident_warning if !@incident
+            return old_incident_warning if result[:incident_id].nil? and @incident.old?
+          end
         end
+
+        run
       end
     end
```
reidmix: I'd make that first `if`-check return to remove some embedding:
```ruby
return unless @match = self.class.regex.match(@message)
```

reidmix: I'm also a fan of `unless`  for the `if !` statements.

lib/chat_ops/command.rb
reidmix: and have the subclasses use the readers

lib/chat_ops/command.rb
```diff
@@ -18,25 +28,34 @@ def help_message(text)
...
-      # retrieve the specified incident from the DB and pass it in as an
-      # argument to run().  If the user does not specify an incident, then
-      # ChatOps.current_incident is used.
-      def parse_incident(should_parse_incident)
-        @should_parse_incident = should_parse_incident
+      # By default, process() tries to parse the incident_id capture group as
+      # an incident, and complains if the incident doesn't exist.  If
+      # the class has incident_optional in its declaration, then the complaint
+      # is suppressed.
+      def incident_optional
```
lexelby: Yeah, I really struggled with this.  I can't put them right together because `incident_optional` is private.  I'll see what I can do.  I wish there were a less-awkward way of doing this.

lib/chat_ops/command.rb
```diff
@@ -51,14 +70,17 @@ def process(user, message)
...
+    #
+    # These utility functions are used by chatops commands.
 
-    def old_incident_warning(incident)
-      ChatOps.message "It looks like you may have forgotten to run `#{Config.chatops_prefix}start incident`.  If you really meant incident #{incident.incident_id}, please specify the incident id with your command."
-    end
+    include ChatOps::MentionHelpers
+    include ChatOps::TimeStampHelpers
+    include ChatOps::MessageHelpers
+    include ChatOps::IncidentHelpers
```
lexelby: Yeah!  I was super-pleased by how this turned out.  No more huge-class-of-death.

lib/chat_ops/command.rb
lexelby: Why would that be preferable?

lib/chat_ops/commands/remove_responder_command.rb
lexelby: I think I'm gonna leave this as it is.  I feel like this way is more explicit.

lexelby:
Added some more cleanup.  This is finally ready for review again.

lib/chat_ops/command.rb
```diff
@@ -1,42 +1,78 @@
...
+      def help
+        [Config.chatops_prefix, @help].compact.join(' ')
       end
+
+      private
+      def setup(&block)
+        s = CommandSetup.new
+        s.instance_eval &block
+        s.config.each do |name, value|
+          instance_variable_set "@#{name}", value
```
reidmix: Nice

lib/chat_ops/command.rb
reidmix: these are my only outstanding comments ^

lib/chat_ops/command.rb
joshuatobin: 👍

lib/chat_ops/command.rb
lexelby: Oops, I'd fixed this but lost it when I reverted a few commits on my dev branch.  Fixed in 491e278.

lexelby:
Okay, I think that's the last of the suggestions.  Ready for a final +1.

PR #69, opened 2016-10-04
Adding chatops timeline command

joshuatobin:
CC: @lexelby @reidmix

### DIFF
--- a/app/models/incident.rb
+++ b/app/models/incident.rb
@@ -17,4 +17,8 @@ class Incident < ActiveRecord::Base
   def chat_end
     super&.in_time_zone Config.time_zone
   end
+
+  def format_timeline
+    timeline_entries.map {|t| ["#{t.timestamp.utc} #{t.user.name}: #{t.message}"]}.flatten.join("\n")
+  end
 end
--- a/app/models/incident.rb
+++ b/app/models/incident.rb
@@ -19,6 +19,6 @@ class Incident < ActiveRecord::Base
   end
 
   def format_timeline
-    timeline_entries.map {|t| ["#{t.timestamp.utc} #{t.user.name}: #{t.message}"]}.flatten.join("\n")
+    timeline_entries.map {|t| "#{t.timestamp.utc} #{t.user.name}: #{t.message}"}.join("\n")
   end
 end
--- a/lib/chat_ops/commands/timeline_command.rb
+++ b/lib/chat_ops/commands/timeline_command.rb
@@ -5,7 +5,7 @@ module ChatOps::Commands
     parse_incident true
 
     def run(user, match, incident)
-      ChatOps.message("Timeline for incident #{incident.incident_id}\n#{incident.format_timeline}")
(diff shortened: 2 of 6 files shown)


app/models/incident.rb
```diff
@@ -17,4 +17,8 @@ def chat_start
   def chat_end
     super&.in_time_zone Config.time_zone
   end
+
+  def format_timeline
+    timeline_entries.map {|t| ["#{t.timestamp.utc} #{t.user.name}: #{t.message}"]}.flatten.join("\n")
```
lexelby: Why the square brackets and the `flatten` call?  Would this work?

`timeline_entries.map {|t| "#{t.timestamp.utc} #{t.user.name}: #{t.message}"}.flatten.join("\n")`

lib/chat_ops/commands/timeline_command.rb
```diff
@@ -0,0 +1,11 @@
+module ChatOps::Commands
+  class TimelineCommand < ChatOps::Command
+    match /timeline(\s+(?<incident_id>\d+))?/
```
lexelby: You'll need to add an end-of-string anchor (`$` or `\Z`).  Otherwise this could match on `h timeline 123 <some message here>`.

lib/chat_ops/commands/timeline_command.rb
```diff
@@ -0,0 +1,11 @@
...
+    help_message "timeline [#] - list the timeline for a incident"
+    parse_incident true
+
+    def run(user, match, incident)
+      ChatOps.message("Timeline for incident #{incident.incident_id}\n#{incident.format_timeline}")
```
lexelby: Don't forget to call `prevent_highlights`.

lib/generators/chatops/chatops_generator.rb
```diff
@@ -4,7 +4,7 @@ class ChatopsGenerator < Rails::Generators::NamedBase
   def copy_file
     create_file "lib/chat_ops/commands/#{file_name}_command.rb", <<-FILE
 module ChatOps::Commands
-  class #{class_name}Command < ChatOps::ChatOpsCommand
+  class #{class_name}Command < ChatOps::Command
```
lexelby: Oops!  My bad, thanks for catching that.

app/models/incident.rb
joshuatobin: Yeah that was dumb. Not sure why I did that. Fixed!

lib/generators/chatops/chatops_generator.rb
joshuatobin: 👍

lib/chat_ops/commands/timeline_command.rb
joshuatobin: @lexelby fixed in 6bdc48c. Also added a spec

lib/chat_ops/commands/timeline_command.rb
joshuatobin: Fixed in b28f6bb

joshuatobin:
@lexelby can I get another look?

lexelby:
:+1: looks great!

PR #70, opened 2016-10-06
Mediator for syncing timeline

joshuatobin:
This is a one-off for migration based purposes. It not idempotent so I could make it a rake task, but either way it will be short lived. 

CC: @lexelby @reidmix

### DIFF
--- a/app/support/mediators/timeline/sync_timeline.rb
+++ b/app/support/mediators/timeline/sync_timeline.rb
@@ -5,7 +5,7 @@ module Mediators::Timeline
       sync_timeline
     rescue Excon::Error
       log_error(error: $!, fn: "call", at: "run")
-      raise $!
+      raise
     end
 
     def sync_timeline
--- a/app/support/mediators/timeline/sync_timeline.rb
+++ b/app/support/mediators/timeline/sync_timeline.rb
@@ -1,0 +1,46 @@
+module Mediators::Timeline
+  class SyncTimeline < Mediators::Base
+
+    def call
+      sync_timeline
+    rescue Excon::Error
+      log_error(error: $!, fn: "call", at: "run")
+      raise $!
+    end
+
+    def sync_timeline
+      data.each do |timeline|
+        # Dealing with erroneous data
+        next unless timeline.second.respond_to?(:each)
+
+        timeline.second.each do |details|
+          d = details.second
+          incident = Incident.find_by(incident_id: d["id"])
+          next if incident.nil?
+
+          responders = d["people"].map {|user| User.find_by(name: user) }
+          incident.responders += responders.compact unless responders.nil?
+
+          d["timeline"].each do |entry|
+            user = User.find_by(name: entry["user"])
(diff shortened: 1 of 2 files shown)


reidmix:
I like it as a mediator.  I agree for a rake task, it can just call the mediator.

app/support/mediators/timeline/sync_timeline.rb
```diff
@@ -0,0 +1,46 @@
...
+      sync_timeline
+    rescue Excon::Error
+      log_error(error: $!, fn: "call", at: "run")
+      raise $!
+    end
+
+    def sync_timeline
+      data.each do |timeline|
+        # Dealing with erroneous data
+        next unless timeline.second.respond_to?(:each)
```
reidmix: Curious as to what is the data when its erroneous.  Could you share a sample of the data?

app/support/mediators/timeline/sync_timeline.rb
```diff
@@ -0,0 +1,46 @@
...
+
+        timeline.second.each do |details|
+          d = details.second
+          incident = Incident.find_by(incident_id: d["id"])
+          next if incident.nil?
```
reidmix: Its surprising to me if a timeline has reference to an Incident that retrodot doesn't know about.
Is this an actual possibility or should it be a `find_or_create_by`?

app/support/mediators/timeline/sync_timeline.rb
```diff
@@ -0,0 +1,46 @@
...
+
+          responders = d["people"].map {|user| User.find_by(name: user) }
+          incident.responders += responders.compact unless responders.nil?
```
reidmix: With no matches, responders would be an empty array `[]` -- I don't think you need `unless responders`

reidmix: Is it a problem if a user cannot be found?

app/support/mediators/timeline/sync_timeline.rb
joshuatobin: Yeah there is some cruft that is coming back as a string vs an Array of timeline data.

app/support/mediators/timeline/sync_timeline.rb
joshuatobin: No there are just so bunk entries in there, possibly from testing.

app/support/mediators/timeline/sync_timeline.rb
joshuatobin: I can remove this. I originally had it in there because incident was coming back nil, again from bunk test data

joshuatobin:
@reidmix TBH I don't think its really worth even creating a rake task to call it. Do the migration, then rip this out.

reidmix:
I guess you could call `runner` to execute it (I'm thinking once its deployed, what steps need to be taken)

joshuatobin:
I'm just going to run it through the console during the migration. `Mediators::Timeline::SyncTimeline.run`.

lexelby:
I'm late to the party, but this looks great.  Sorry for the junk data!

PR #71, opened 2016-10-06
App json fix

joshuatobin:


joshuatobin:
@reidmix @lexelby Merging this. The error was being masked by the apps-setup API.

PR #72, opened 2016-10-07
Generate Retrospective Email 

joshuatobin:
This PR allows sending email via ActiveAdmin using activemailer. I've setup a simple activemailer haml template that will be used for the MVP. We'll save editing the email before sending for another sprint.

retrodot-email

CC: @reidmix @lexelby


app/admin/incident.rb
```diff
@@ -36,14 +36,29 @@ class TrelloAuthRequired < StandardError; end
...
+        redirect_to collection_path, flash: { error: "Error: #{e}" }
     else
-      redirect_to collection_path, notice: "Documents for Retrospective '#{incident.title}' have been prepared."
+      redirect_to collection_path, notice: "Documents for Retrospective '#{resource.title}' have been prepared."
     end
   end
 
   member_action :send_email, method: :post do
-    redirect_to collection_path, notice: "PLACEHOLDER: will send email: #{params['inputs']}"
+    if resource.trello_url.nil? || resource.google_doc_url.nil?
```
reidmix: Can this be a method on "resource"? Meaning something like `resource.retro_prepared?`
I like when its descriptive, and if there are other changes to what `retro_prepared?` means it's centralized.

app/admin/incident.rb
```diff
@@ -36,14 +36,29 @@ class TrelloAuthRequired < StandardError; end
...
+    end
+
+    email = JSON.parse(params['inputs'])
+
+    NotificationsMailer.retro_followup(
+      incident: resource,
+      sender:   current_user.email,
+      to:       email["To"],
+      cc:       email["CC"],
+      subject:  email["Subject"]
```
reidmix: Can any of these be determined automatically or pre-populated in the form (to, cc, subject)?
Especially subject since it's always the same when I send out the emails today.

config/config.rb
```diff
@@ -47,4 +47,11 @@ module Config
   # https://developers.google.com/apps-script/execution
   optional :google_script_id, string
   optional :google_script_function, string
+
+  # Mailer
+  optional :mailgun_smtp_server, string
+  optional :mailgun_smtp_port, string
+  optional :mailgun_domain, string
+  optional :mailgun_smtp_login, string
+  optional :mailgun_smtp_password, string
```
reidmix: We could drop the `mailgun_` on these names to keep it generic. (and use `smtp_domain` for `mailgun_domain`)

Gemfile
```diff
@@ -34,6 +34,7 @@ gem 'ruby-trello', "~> 1.5.1"
 gem 'default_value_for', '~> 3.0.0'
 gem 'google-api-client', '~> 0.9.15'
 gem 'googleauth'
+gem 'mailgun_rails'
```
reidmix: I'm somewhat sure `mailgun_rails` isn't needed because mail is getting configured with the built-in rails configurations.  This is how we do it on `heroku/status` too.   Plus we don't need the custom variables or headers.

app/admin/incident.rb
joshuatobin: Great idea. Added in 35c06a5

config/config.rb
joshuatobin: Good call. Fixed in a05cad3

Gemfile
joshuatobin: Yeah this was cruft

app/admin/incident.rb
```diff
@@ -36,14 +36,29 @@ class TrelloAuthRequired < StandardError; end
...
+        redirect_to collection_path, flash: { error: "Error: #{e}" }
     else
-      redirect_to collection_path, notice: "Documents for Retrospective '#{incident.title}' have been prepared."
+      redirect_to collection_path, notice: "Documents for Retrospective '#{resource.title}' have been prepared."
     end
   end
 
   member_action :send_email, method: :post do
-    redirect_to collection_path, notice: "PLACEHOLDER: will send email: #{params['inputs']}"
+    if resource.retro_prepared?
```
reidmix: <3

app/admin/incident.rb
reidmix: it is -- look down on lines 89-91

app/models/incident.rb
```diff
@@ -21,4 +21,8 @@ def chat_end
   def format_timeline
     timeline_entries.map {|t| "#{t.timestamp.utc} #{t.user.name}: #{t.message}"}.join("\n")
   end
+
+  def retro_prepared?
+    trello_url.nil? || google_doc_url.nil?
```
reidmix: i think you want `trello_url && google_doc_url`

app/admin/incident.rb
reidmix: s/if/unless

app/models/incident.rb
joshuatobin: Yeah thats clearer. thx

joshuatobin: @reidmix fixed in 062c2b2

PR #73, opened 2016-10-08
ChatOps API Updates

joshuatobin:
A few updates to the chatops API. Added `require_all` to the controller so that the `commands` Array would be populated and some other minor changes. 

CC: @reidmix @lexelby

### DIFF
--- a/app/controllers/api/v1/chat_ops_controller.rb
+++ b/app/controllers/api/v1/chat_ops_controller.rb
@@ -26,5 +26,8 @@ class Api::V1::ChatOpsController < ApplicationController
     params.require(:message)
   end
 
-  require_all 'lib'
+  # Load all .rb files in the 'commands' subdirectory using the require_all gem.
+  # Rails's autoloading won't automatically load the subclasses, so without this
+  # ChatOps.commands would return an empty array.
+  require_all 'lib/chat_ops/commands/**.rb'
 end
--- a/app/controllers/api/v1/chat_ops_controller.rb
+++ b/app/controllers/api/v1/chat_ops_controller.rb
@@ -7,7 +7,7 @@ class Api::V1::ChatOpsController < ApplicationController
   end
 
   def responder
-    user = User.ensure(user_params)
+    user   = User.ensure(user_params)
     result = ChatOps.process(user, message_params)
 
     if result
--- a/app/controllers/api/v1/chat_ops_controller.rb
+++ b/app/controllers/api/v1/chat_ops_controller.rb
@@ -6,7 +6,7 @@ class Api::V1::ChatOpsController < ApplicationController
     render json: {"regex": ChatOps.matcher}
   end
 
-  def responder
+  def respond
     user   = User.ensure(user_params)
     result = ChatOps.process(user, message_params)
 
--- a/app/controllers/api/v1/chat_ops_controller.rb
+++ b/app/controllers/api/v1/chat_ops_controller.rb
@@ -6,9 +6,9 @@ class Api::V1::ChatOpsController < ApplicationController
     render json: {"regex": ChatOps.matcher}
   end
 
-  def respond
-    user = User.ensure(**params.permit(user: [:email, :name, :handle])[:user].to_h.symbolize_keys)
-    result = ChatOps.process(user, params['message'])
+  def responder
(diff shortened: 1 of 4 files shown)


app/controllers/api/v1/chat_ops_controller.rb
```diff
@@ -6,14 +6,28 @@ def matcher
     render json: {"regex": ChatOps.matcher}
   end
 
-  def respond
-    user = User.ensure(**params.permit(user: [:email, :name, :handle])[:user].to_h.symbolize_keys)
-    result = ChatOps.process(user, params['message'])
+  def responder
+    user   = User.ensure(user_params)
+    result = ChatOps.process(user, message_params)
```
reidmix: nice!

app/controllers/api/v1/chat_ops_controller.rb
```diff
@@ -6,14 +6,28 @@ def matcher
...
     if result
       render json: result
     else
       render status: :not_found, json: {"error": "command not found"}
     end
   end
+
+  private
+  def user_params
+    params.require(:user).permit(:email, :name, :handle).to_h.symbolize_keys
```
reidmix: I dont *think* you need the `to_h.symbolize_keys` anymore.

joshuatobin: Yeah thats originally what I thought but it didn't work. Give me a few and I'll provide an example.

app/controllers/api/v1/chat_ops_controller.rb
joshuatobin: @reidmix Also if you have idea for better naming for the endpoint. I'm still not 100% on response/responder

app/controllers/api/v1/chat_ops_controller.rb
joshuatobin: @reidmix Ah yes. `require` returns an instance of `ActionController::Parameters`. Which is why I had to do this to make it work. 

`<ActionController::Parameters {"email"=>"tobin@heroku.com"} permitted: true>`

app/controllers/api/v1/chat_ops_controller.rb
reidmix: Yeah, typically actions are verbs: "ChatOps responds" kinda thing.   
What's happening is the request is processed or handled, so `process`, `handle`, `respond` (which is what it was) is where I land -- maybe more specifically `handle_command`, `process_message` might be more along the lines you're thinking?

app/controllers/api/v1/chat_ops_controller.rb
reidmix: right right right -- i forgot.

you might get away with not using `symbolize_keys` (name parameters take string keys).

lexelby: I wrote this with `symbolize_keys` because I was passing in the params hash as keyword args to `User.ensure` using `**`.  For `**`, ruby requires a hash with keys that are symbols.  Maybe the `**` isn't needed?

app/controllers/api/v1/chat_ops_controller.rb
joshuatobin: Okay I left this as `respond`

app/controllers/api/v1/chat_ops_controller.rb
joshuatobin: this is still needed. 

```
ArgumentError (wrong number of arguments (given 1, expected 0)):

app/models/user.rb:17:in `ensure'
```

joshuatobin:
@lexelby @reidmix Can I get one last review here?

app/controllers/api/v1/chat_ops_controller.rb
reidmix: You're right -- my testing of it was bad, hahah, I was doing:
```ruby
test "a": 1, "b": 2, "c": 3
```
which of course converts them to symbols. 
```ruby 
test "a" => 1, "b" => 2, "c" => 3
``` 
Does not work, as you showed.  

Sorry bout that!

joshuatobin: 👍

PR #75, opened 2016-10-12
loggerator update

reidmix:
Includes update to initializer, because new version has less name collisions.

/cc @lexelby @joshuatobin


joshuatobin:
👍

joshuatobin:
@reidmix this made it in via another PR.

PR #76, opened 2016-10-14
Add support for chatops via slack slash commands

joshuatobin:
This PR introduces running chatops commands from [slack via slash commands](https://api.slack.com/slash-commands). The reason why slash commands are being used over a traditional bot is mostly one of security. We provide a simple endpoint that is validated by a slack slash token. Using a bot would require us to provide a token that has access to all channels and chat history etc. Slash commands are much simpler. 

Some things to note about this implementation:
* Using ActiveJob alongside sidekiq to queue command requests. Slack requires a response within 3000ms, so we send back a 201, then queue the followup response.

* Added a commands controller to validate a `current_user`s slack `user_id` along with the slack token which then queues to ActiveJob. This should allow us to keep the endpoint fairly generic and just pass in simple regexs to `ChatOps.process`.

* Unfortunately the `omniauth-slack` gem is in a bit of limbo. We require `signin with slack` support which is not yet in master. For now I've forked the repo and pointed the Gemfile at my fork. This isn't ideal but I've pinged the maintainer and will see if we can make any progress on that front. 

* We may need to do some regex massaging on the `timeline` command. Right now to run timeline the command would be `/timeline timeline 952`, vs just accepting `952` as the argument.

*Needs documentation on setting up the slack oauth. 

CC: @reidmix @lexelby @ys @atmos


cmeckhardt:
We've had a bunch of problems getting used to "h timeline", both on SRE and in the wider engineering org.  Do we have a transition plan for this change that's not just replacing "h timeline" with this one day?  I'm concerned if we swap this over without a gradual transition that it will impact our ability to document incidents and make retros harder.

joshuatobin:
The transition plan will consist of us letting the eng org know, documenting it, and then replacing the existing tooling with a message the directs the user to /timeline.

app/controllers/auth_controller.rb
```diff
@@ -27,8 +27,14 @@ def callback
...
 
-    origin = session.delete(:return_to)
-    redirect_to (origin && origin !~ %r{/auth/} ? origin : root_path)
+    redirect
+  end
+
+  def slack_callback
+    update_slack = User.from_omniauth(omniauth_info)
+
+    redirect_to failure_auth_path and return unless update_slack
```
lexelby: Neat!  I like this syntax.

cmeckhardt:
Based on the issues we've had with "h timeline" adoption, I don't think that's enough.  Can we investigate having an overlap period?

app/models/user.rb
```diff
@@ -14,6 +15,15 @@ def as_json(options={})
   end
 
   class << self
+
+    def from_omniauth(omniauth_info)
```
lexelby: I feel like this method should have "slack" in its name, since it's slack-specific.

lib/chat_ops/chat_ops.rb
```diff
@@ -150,7 +150,7 @@ def normalize_name(name)
 
     def find_and_remove_names!(message)
       names_to_users = User.pluck(:name, :id).each_with_object({}) do |(name, id), h|
-        h[normalize_name(name)] = id
+        h[normalize_name(name)] = id unless name.nil?
```
lexelby: Perhaps instead add a scope, `User.with_names`?  In fact, I could have sworn @reidmix suggested that at some point and that I implemented it... /shrug

lib/chat_ops/commands/end_incident_command.rb
```diff
@@ -8,7 +8,7 @@ def run(user, match, incident)
       incident.chat_end = Time.now
       incident.save
 
-      return { message: "Recorded the end of chat for incident #\#{incident.incident_id} at #{incident.chat_end.inspect}." }
+      return { message: "Recorded the end of chat for incident: #{incident.incident_id} at #{incident.chat_end.inspect}." }
```
lexelby: I don't really feel like a colon makes sense here...

app/controllers/commands_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+      render json: { response_type: "in_channel" }, status: 201
+    else
+      render json: set_slack_token, status: 202
+    end
+  end
+
+  private
+
+   def current_user
+     @current_user || User.find_by(slack_external_id: command[:user_id])
```
lexelby: What is this trying to do with `@current_user`?  I think you might have meant `||=`, but I don't understand why you would want to save the current user between commands...

app/controllers/commands_controller.rb
```diff
@@ -0,0 +1,37 @@
+class CommandsController < ApplicationController
+  skip_before_action :verify_authenticity_token, only: :create
+
+  def create
+    return render json: {}, status: 403 unless valid_slack_token?
+
+    if current_user
+      SlackCommandJob.perform_later(current_user, command)
+      render json: { response_type: "in_channel" }, status: 201
```
lexelby: Why code 201 and 202 here?  The [slash-command documentation](https://api.slack.com/slash-commands) says:

> Your URL should respond with a HTTP 200 "OK" status code. Any other flavor of response will result in a user-facing error.

app/controllers/commands_controller.rb
lexelby: As this is written, it requires that every user that wants to use /timeline must do the oauth dance.  However, I'm pretty sure we can avoid that.  The way beeper does it is to have _one_ member of the tools team oauth with slack, and request scopes that allow beeper to use the slack REST API to look up users.  Then when a command arrives, it sends off a request to look up the slack user ID and ask slack for their email address.  We could do that here and avoid having every user have to oauth to slack.

lexelby:
I second @cmeckhardt, and I'd also point to the issues we continue to have with `/pd` vs `h page`.

reidmix:
Traditionally, I know the DJ has been very buggy and created problems for many groups and they've generally moved off of DJ in favor of Sidekiq, or just use Sidekiq to start.   For that reason, I'm hesitant to use DJ. 

Additionally, we've been using Sidekiq everywhere which makes it easy for us to go from one app to another and understand the same semantics.   From what I've learned recently, ActionCable might make that even more irrelevant (since it's a layer on top of the async systems).

I find Sidekiq to be very simple as well, what would you say are the advantages to DJ?

app/controllers/commands_controller.rb
joshuatobin: General mistake.

app/controllers/commands_controller.rb
joshuatobin: @ys @atmos - Any thoughts on the oauth implementation here?

lib/chat_ops/chat_ops.rb
joshuatobin: Good call 👍

app/models/user.rb
joshuatobin: sure.

app/controllers/auth_controller.rb
joshuatobin: Yeah I wasn't sure but it reads quite nicely so 😋

app/controllers/commands_controller.rb
ys: For Beeper, I have one user installing it and then find_or_create based on the slack user id.
If you want to do the oauth dance just for that... It is not useful to have them to go to the server.

joshuatobin:
Update: I'm going to rip out the async stuff for now and keep it simple. I'm also going to use a different scope so we receive back just one access token which will not require everyone to log in.

lexelby:
@reidmix just to close the loop, I'll answer your question anyway: DJ seems simpler to set up (no need to set up redis and whatnot) and doesn't require paying for a redis instance.

app/controllers/auth_controller.rb
```diff
@@ -27,8 +27,14 @@ def callback
 
     update_google_creds unless developer?
 
-    origin = session.delete(:return_to)
-    redirect_to (origin && origin !~ %r{/auth/} ? origin : root_path)
+    redirect
```
reidmix: Ha. In prior companies we had a similar thing and called it `redirect_back` or `redirect_back_or_default`. 
I feel like I've come full circle :D

app/jobs/slack_command_job.rb
```diff
@@ -0,0 +1,30 @@
+class SlackCommandJob < ApplicationJob
+  queue_as :default
+
+  def perform(user, message)
+    result = ChatOps.process(user, message[:text])
+
+    if result
+      result = result[:reaction] ? result[:reaction] : result[:message]
+    else
+      no_result
```
reidmix: This is not doing anything -- i'm thinking you may want:
```ruby
result = if result ... else ... end
```

app/jobs/slack_command_job.rb
```diff
@@ -0,0 +1,30 @@
...
+    end
+
+    resp = {
+      text: result,
+      response_type: "in_channel"
+    }
+
+    Excon.post(message[:response_url], body: resp.to_json, headers: headers)
+  rescue Excon::Error
+    log_error($!)
```
reidmix: When 0.0.2 of loggerator is merged, you just need `log_error`

app/models/user.rb
```diff
@@ -14,6 +16,15 @@ def as_json(options={})
   end
 
   class << self
+
+    def from_slack_omniauth(omniauth_info)
+      user = find_by(email: omniauth_info["info"]["email"] )
+      return if user.nil?
```
reidmix: Would you want to create the user at this point?

config/routes.rb
```diff
@@ -6,6 +6,10 @@
 
   get '/auth/trello/callback', to: 'trello#create'
 
+  # /slash command endpoint for slack
+  post "/slack/command",       to: "commands#create"
+  get  "/auth/slack/callback", to: "auth#slack_callback"
```
reidmix: Isn't this covered by lines 13-14 below?  
It would simplify the routes if they were.

reidmix: Then `auth#callback` then could just call the provider specific callback, which would keep each auth method focused on the provider.

app/controllers/commands_controller.rb
atmos: There's two different ways to manage this. There's `installation` and `identify`. Installation is for installing the application and making the slash command available. Identify is for verifying that a user is a member of a specific slack channel.

The installation of the slash command onto your team ensures that people triggering the commands are doing so from inside of a channel on your slack team. I can't recall if the slash commands give you things like email addresses, but if they do you probably don't need identify from them.

The identify stuff ends up being useful if you have actions going on outside of the slash-command workflow. So we use it in things like rolodex where a user goes from chat, to a web page, and then is verified against other services outside of the slack context.

app/models/user.rb
```diff
@@ -4,7 +4,9 @@ class User < ApplicationRecord
   validates :handle, format: { with: /\A[-_.a-z0-9]+\z/, message: "only allows numbers, lowercase letters, dashes, periods and underscores" }, allow_nil: true
 
   scope :with_handle, -> { where.not(handle: [nil, ""]) }
+  scope :with_name,   -> { where.not(name: [nil, ""]) }
 
+  # TODO: Move to active model serializer
```
atmos: Fwiw we found a method returning a hash to be a lot simpler than the overhead of active model serializer.

app/models/user.rb
atmos: It depends on how the user is seeded initally, but I can't tell where that is based on this diff.

config/initializers/omniauth.rb
```diff
@@ -11,4 +11,5 @@
                 https://www.googleapis.com/auth/script.external_request]
     }
   provider :trello, Config.trello_consumer_key, Config.trello_consumer_secret, app_name: Config.app_name, scope: "read,write,account", expiration: "never"
+  provider :slack, Config.slack_client_id, Config.slack_client_secret, scope: "identity.basic,identity.email"
```
atmos: I believe the scopes you're going to need here is `install`. You'll only do one slack installation for the org, but that's how the slash-commands will be exposed.

[Reference Routes](https://github.com/heroku/beeper/blob/master/config/routes.rb#L9-L10) [Reference OmniAuth setup](https://github.com/heroku/beeper/blob/master/config/initializers/omniauth.rb#L2-L3)

config/routes.rb
atmos: We ended up with different, explicit, [callback handlers](https://github.com/heroku/rolodex/blob/master/app/controllers/sessions_controller.rb#L11-L55) in multi-provider environments.

I think it's a matter of what your team is more comfortable with really.

app/models/user.rb
reidmix: Right, now that I know a little more, it sounds like the install will happen, then from that point on they will find or create the user after that, AFAICT.  

Your comment does point out an interesting fact that normally the (google oauth) lookup happens in the `AuthController#callback` method.   For me, it would be nice that they were handled consistently.

config/routes.rb
reidmix: That looks nice.  👍

Procfile
```diff
@@ -1,2 +1,3 @@
 web:     bundle exec puma -t ${MIN_THREADS:-2}:${MAX_THREADS:-6} -w ${WEB_WORKERS:-2} -p ${PORT:-3000} -e ${RACK_ENV:-development}
+worker:  bundle exec sidekiq -c ${SIDEKIQ_THREADS:-3} -e ${RACK_ENV:-development}
```
ys: You might want to rely on `RAILS_MAX_THREADS` which is the default for tuning DB concurrency in rails already.

See https://github.com/mperham/sidekiq/issues/2985 for more info by @schneems

app/controllers/auth_controller.rb
```diff
@@ -53,13 +59,13 @@ def verify
   end
 
   def developer?
-    request.env['omniauth.auth']['provider'] == "developer"
+    omniauth_info['provider'] == "developer"
```
ys: Do you have conventions around quotes internally? This bugs me.

app/controllers/auth_controller.rb
```diff
@@ -71,4 +77,13 @@ def auth_provider
   def valid_email? email
     !!(email =~ %r[@#{Config.google_domain}$])
   end
+
+  def redirect
+    origin = session.delete(:return_to)
+    redirect_to (origin && origin !~ %r{/auth/} ? origin : root_path)
```
ys: You might want to get just a path here.
Where is `return_to` set? It might be a full url and will redirect you to another website

`https://myphishyretrodot.com/auth/slack`

app/controllers/commands_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+  def slack_token
+    Config.slack_slash_command_token
+  end
+
+  def command
+    params.permit(:text, :token, :user_id, :response_url, :command).to_h.symbolize_keys
+  end
+
+  def set_slack_token
+    { text: "Please set slack token at: #{Config.frontend_host}/auth/slack", response_type: "in_channel" }
```
ys: Isn't it a config var?

app/controllers/auth_controller.rb
reidmix: On our last project we did all double-quotes, which is my preference, since the performance difference is negligible.   Agreed, we should stick to a style.  RE consider starting using rubocop.

ys: You can steal our rubocop config if you want, it helps not bikeshedding and focus:D

joshuatobin: thanks @ys. I'll do that. cheers.

joshuatobin:
Closing in favor of https://github.com/heroku/retrodot/pull/80

PR #77, opened 2016-10-18
legacy bot support

lexelby:
A change and bugfix to support our legacy chatops bot.


app/models/user.rb
```diff
@@ -4,6 +4,7 @@ class User < ApplicationRecord
   validates :handle, format: { with: /\A[-_.a-z0-9]+\z/, message: "only allows numbers, lowercase letters, dashes, periods and underscores" }, allow_nil: true
 
   scope :with_handle, -> { where.not(handle: [nil, ""]) }
+  scope :with_name, -> { where.not(name: [nil, ""]) }
```
reidmix: It'd be really nice to make all missing values `nil` then you don't have to test/scope for the empty string `""`. 
Can we normalize the `name`/`handle` when inserting into the db?

lib/chat_ops/chat_ops.rb
```diff
@@ -15,11 +15,10 @@ def commands
...
 
     # Generates a regex that matches any of the declared ChatOps::Command
-    # subclassess.  Each ChatOps::Command subclass has its own regex, so we use
-    # Regexp.union to combine them.  It's almost the same as joining them with
-    # '|'.
+    # subclassess.  We massage the regex a bit so that it's more palatable to
+    # regex implementations with fewer features than Ruby's.
     def matcher
-      Regexp.union(commands.collect(&:regex))
+      "(?ix)" + commands.map(&:regex).map(&:source).join('|').gsub(/\(\?<[^>]+>/, '(')
```
reidmix: can you help me understand why we are getting rid of named group `(?<thing>` parts of regex?
I guess i don't understand the intent.

reidmix: Also, could you help me understand why we have to make it "more palatable to regex implementations with fewer features than Ruby's"?  I'm guessing these are related :D

reidmix: Oh, I think I get it, `matcher` is for use outside Retrodot (presumably heimdall), and the named groups are making it puke.  Is that right?  If so, can we put a comment that matcher is intended for external use?

app/models/user.rb
lexelby: I'll look into this.

lib/chat_ops/chat_ops.rb
lexelby: Yup, you got it.  I'll clarify the comment.

joshuatobin:
👍

PR #78, opened 2016-10-18
Adding omniauth-slack support

joshuatobin:
Adds support for installing slack and adding an oauth token. The token will be attached to one user and will make requests "on behalf of" other users. 

CC: @reidmix @lexelby

### DIFF
--- a/app/controllers/auth_controller.rb
+++ b/app/controllers/auth_controller.rb
@@ -58,19 +58,12 @@ class AuthController < ApplicationController
 
   def install_slack
     redirect_to failure_auth_path and return unless current_user
+    current_user.update(slack_access_token: omniauth_info["credentials"]["token"])
 
-    update_slack_token
-    flash[:notice] = "Updated slack token"
-
-    redirect_to root_path
+    redirect_to root_path, notice: "Updated slack token"
   end
 
   protected
-
-  def update_slack_token
-    current_user.update(slack_access_token: omniauth_info["credentials"]["token"])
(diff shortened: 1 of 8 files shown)


app/controllers/auth_controller.rb
```diff
@@ -45,22 +45,36 @@ def unauthorized
...
+
+    update_slack_token
+    flash[:notice] = "Updated slack token"
+
+    redirect_to root_path
   end
 
   protected
 
+  def update_slack_token
```
joshuatobin: heh. Because in the past alot of times in the past @reidmix would have me move it into its own method.

reidmix: i'd like moving complicated code or not-obvious code to a single method when giving it a name helps give intent (or when it is repeated).  It's more of a case by case basis.  In this case it's fairly straightforward and I'd say it can be inline.

joshuatobin:
@lexelby @reidmix made suggested changes

awenten:
Not sure why it didn't remove her from the repo but I do see that she is
still on it.


*Alex Wenten*
Senior Helpdesk Analyst, Vibe Operations
awenten@heroku.com
Heroku / Salesforce
http://www.heroku.com

awenten:
Please Disregard the last email.


*Alex Wenten*
Senior Helpdesk Analyst, Vibe Operations
awenten@heroku.com
Heroku / Salesforce
http://www.heroku.com

PR #79, opened 2016-10-18
Adding slack sync mediator

joshuatobin:
This adds a mediator for syncing slack users. The reasoning behind this is to keep the user information up to date so that `ChatOps.process` doesn't have rely on looking up this info for each command request. 

CC: @lexelby @reidmix


app/support/mediators/user/sync_slack_users.rb
```diff
@@ -0,0 +1,26 @@
...
+        ::User.ensure(**user.slice(:email, :name, :handle))
+      end
+    end
+
+    private
+
+    def users
+      @client.users_list["members"].map do |u|
+        [handle: u["handle"], name: u["name"], email: u["profile"]["email"]]
+      end.flatten
```
lexelby: Why are you putting each hash into an array on line 21 and then flattening back to a single array on line 22?  Why not just have this on line 21?

`{handle: u["handle"], name: u["name"], email: u["profile"]["email"]}`

app/support/mediators/user/sync_slack_users.rb
```diff
@@ -0,0 +1,26 @@
...
+  class SyncSlackUsers < Mediators::Base
+
+    def initialize(token:)
+      @client = Slack::Client.new token: token
+    end
+
+    def call
+      log action: 'syncing'
+      users.each do |user|
+        ::User.ensure(**user.slice(:email, :name, :handle))
```
lexelby: Since you're building the user hash in the `users` method, you don't need to `slice` here, you can just use it directly.

joshuatobin: 👍

joshuatobin:
@lexelby Addressed comments.

app/support/mediators/user/sync_slack_users.rb
```diff
@@ -10,16 +10,16 @@ def initialize(token:)
     def call
       log action: 'syncing'
       users.each do |user|
-        ::User.ensure(**user.slice(:email, :name, :handle))
+        ::User.ensure(user) unless user[:email].nil?
```
lexelby: Oh, good catch.  I wonder if `.blank?` might be better here just in case slack does something weird?

joshuatobin: +1 good call

joshuatobin:
@lexelby can you look again. I also realized we need the `user_id` field from slack. This is the only mapping we'll have from the slash command payload.

joshuatobin:
Needs a rebase :/

joshuatobin:
Ready for review again.

lexelby:
Looks great!  Nice work.

reidmix:
Agreed, this is quite nice 👍

PR #80, opened 2016-10-20
Add slack support for chatops

joshuatobin:
CC: @lexelby @reidmix


app/controllers/messages_controller.rb
```diff
@@ -0,0 +1,32 @@
+class MessagesController < ApplicationController
+  skip_before_action :verify_authenticity_token, only: :create
+
+  def create
+    return render_error("Invalid Token", 401) unless valid_slack_token?
+
+    user = User.slack_user_lookup(message[:user_id])
+    return render_error("Slack `user_id` not found in retrodot", 403) unless user
+
```
joshuatobin: I may try and refactor these returns a bit more. Still trying to figure out a cleaner way to get through all the unless logic.

app/models/message.rb
```diff
@@ -0,0 +1,23 @@
...
+      result = ChatOps.process(user, message[:text])
+
+      parse_result(result)
+    end
+
+    private
+
+    def parse_result(result)
+      result = if result
+        result[:reaction] ? result[:reaction] : result[:message]
```
lexelby: Are you thinking we'll update this later to handle the more complicated message formatting, and see if doing an actual reaction is possible?  If not, we could just modify the commands to return simple messages instead?

lib/chat_ops/commands/end_incident_command.rb
```diff
@@ -9,7 +9,7 @@ def run
       @incident.chat_end = Time.now
       @incident.save
 
-      return { message: "Recorded the end of chat for incident #\#{@incident.incident_id} at #{@incident.chat_end.inspect}." }
+      return { message: "Recorded the end of chat for incident #{@incident.incident_id} at #{@incident.chat_end.inspect}." }
```
lexelby: Does this break slack formatting or something?

app/models/message.rb
```diff
@@ -0,0 +1,23 @@
...
+      else
+        "No results found."
```
lexelby: perhaps something like "command unknown, try `/timeline help` for help"?

spec/controllers/messages_controller_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+    }
+  }
+
+  before do
+    allow(Config).to receive(:slack_slash_command_token).and_return("123456789")
+  end
+
+
+  describe "#Create" do
+    it "processes a message" do
```
lexelby: Perhaps this test should mock `ChatOps.process` so that we're not testing the `start incident` command itself?  Also you could ensure that `ChatOps.process` is called once.

app/controllers/messages_controller.rb
lexelby: It's pretty readable as it stands now, FWIW.

lib/chat_ops/commands/end_incident_command.rb
joshuatobin: Yup

app/models/message.rb
joshuatobin: I originally started down the path of refactoring `command` to return differently. I can do that in a separate PR if thats a road we want to go down? Just trying to keep it as simple as possible at first but im happy to change it.

spec/controllers/messages_controller_spec.rb
joshuatobin: 👍

app/controllers/auth_controller.rb
```diff
@@ -56,9 +57,10 @@ def developer?
     omniauth_info["provider"] == "developer"
   end
 
+  # Requires user to be logged in before fetching the token.
   def install_slack
-    redirect_to failure_auth_path and return unless current_user
-    current_user.update(slack_access_token: omniauth_info["credentials"]["token"])
+    current_user.update(slack_access_token: slack_access_token)
```
reidmix: Are we sure there's always a `slack_access_token`?
What happens when this method is called the second time, will we have two `slack_access_token`s?
What happens in that case?

app/controllers/messages_controller.rb
```diff
@@ -0,0 +1,32 @@
...
+    result = Message.process_message(user, message)
+
+    render json: { text: result, response_type: "in_channel" }, status: 200
+  end
+
+  private
+
+  def valid_slack_token?
+    ActiveSupport::SecurityUtils.secure_compare(message[:token], slack_token) rescue false
```
reidmix: Doesn't look like `secure_compare` throws any exceptions from what I can tell.

app/models/message.rb
```diff
@@ -0,0 +1,23 @@
+class Message < ActiveRecord::Base
```
reidmix: Doesn't seem this is a model (ie. no corresponding db table) -- you can drop the inheritance (maybe make it a mediator?)

app/models/message.rb
reidmix: This is a weird one -- I guess I'd expect the formatting to be taken care of `ChatOps.process`  -- but I also kinda like that the processing and formatting are separate.  Is the formatting going to be much more complicated?  If not, I think I would go ahead and fold it into  `ChatOps.process`

app/controllers/messages_controller.rb
```diff
@@ -0,0 +1,32 @@
+class MessagesController < ApplicationController
+  skip_before_action :verify_authenticity_token, only: :create
+
+  def create
+    return render_error("Invalid Token", 401) unless valid_slack_token?
+
+    user = User.slack_user_lookup(message[:user_id])
```
reidmix: Might be worth just having the `find_by` here since its so straightforward.

config/config.rb
```diff
@@ -58,4 +58,5 @@ module Config
 
   optional :slack_client_id, string
   optional :slack_client_secret, string
+  optional :slack_slash_command_token, string
```
reidmix: maybe just `slack_token` since its called that in a method earlier?

app/controllers/auth_controller.rb
reidmix: Also, where else does this get used (read from)?  Is that an upcoming PR? :D

reidmix: I'm also pro 👍 moving this into a `Team` model when you get to/refactor it -- it will remove some of these questions above.

app/controllers/messages_controller.rb
joshuatobin: fair enough.

app/controllers/messages_controller.rb
joshuatobin: If you pass it nil it does:

```
ActiveSupport::SecurityUtils.secure_compare('hi', nil)
NoMethodError: undefined method `bytesize' for nil:NilClass
```

app/models/message.rb
lexelby: My idea was that commands could use helper functions on `Command` to return formatted messages, and then the glue code in `Message` (or wherever) could do whatever is necessary to get the formatted message sent to Slack (or another chatops provider?) properly.

I'd really like to try to keep parity with the formatting done by the current chatops stuff.  That means that the timeline should be formatted as a slack attachment so that it starts out collapsed.  Otherwise, it'll spam the channel during an incident.  I think we may not be able to do the reactions thing, which I'll miss, but the alternative of a message "only you can see" saying that their command was received is probably just as good.

spec/controllers/messages_controller_spec.rb
joshuatobin: Added in e31cfb0

app/controllers/auth_controller.rb
joshuatobin: @reidmix So the idea here is that this will be a one time operation. I'm going to add some code to the admin UI with a link and message instructing the user what to do. That should only get called once but yes, if the method is invoked a second time we'll have two tokens. I was considering adding a scope here to find the token. I think what I'm going to do is add some logic to allow only one token to be set.

app/controllers/messages_controller.rb
joshuatobin: 7ecb552

app/models/message.rb
joshuatobin: okay sounds good

config/config.rb
joshuatobin: @reidmix hrm. I purposely named it like this since its not obvious what kinda of slack token it is. I feel like being verbose here is a good thing.

app/models/message.rb
lexelby: BTW, I think this is equivalent:

`result[:reaction] || result[:message]`

or even

`result[:reaction] or result[:message]`

app/controllers/auth_controller.rb
joshuatobin: @reidmix See 45818f4. I added some logic to allow only for one token using scopes.

reidmix: Once the token gets set, where does it get used?

joshuatobin:
@lexelby @reidmix Comments addressed. Please give another looksy

config/config.rb
reidmix: OK, maybe rename the `slack_token` method that wraps it in MessagesController, so there's no confusion?

app/controllers/auth_controller.rb
joshuatobin: The token will be used in the `SyncSlackUsers` mediator, which still needs to be updated.

app/controllers/auth_controller.rb
```diff
@@ -59,13 +59,19 @@ def developer?
 
   # Requires user to be logged in before fetching the token.
   def install_slack
-    current_user.update(slack_access_token: slack_access_token)
-    log(fn: :install_slack, at: :update_slack_token, user: current_user.name)
+    return update_slack_token if User.with_slack_token.empty?
```
lexelby: Hmm, I don't think rejecting a second token is the right way to go.  Let's say you're the person who does the slack oauth dance and gets your token saved.  Then you have to do a slack credroll or whatever, and you invalidate all oauthed apps in your slack account.  Your token is now invalid, so we need to replace it in retrodot.  If you oauth again (or I oauth) the new token should replace the existing one.

joshuatobin:
Added one more commit for warning the users if the token isn't set: 67e651c

app/controllers/messages_controller.rb
```diff
@@ -4,22 +4,20 @@ class MessagesController < ApplicationController
   def create
     return render_error("Invalid Token", 401) unless valid_slack_token?
 
-    user = User.find_by(message[:user_id])
+    user = User.find_by(slack_user_id: message[:user_id])
     return render_error("Slack `user_id` not found in retrodot", 403) unless user
```
lexelby: Oh!  Maybe you can condense is to:

`user = User.find_by(slack_user_id: message[:user_id]) or return render_error(...)`

Not sure if that's more readable, though...

app/controllers/auth_controller.rb
joshuatobin: All you would have to do is just nil out the token. This can all go away when we build a teams controller.

lib/chat_ops/commands/remove_responder_command.rb
```diff
@@ -11,7 +11,7 @@ def run
 
       @incident.responders -= responders
 
-      reaction(':checkmark:')
+      message(':checkmark:')
```
lexelby: If we're not able to do reactions, I'd really rather not just respond with the emoji.  Single emoji like this become "jumboji" -- big images.  It's pretty disrupting, which is why I did the reaction thing in heimdall in the first place.

Can we instead send an acknowledgement message only to the person that did the /command, and otherwise be silent?

joshuatobin:
@reidmix So you want to see the message class live in `Api::V1::ChatOpsController` ?

joshuatobin:
@reidmix @lexelby  I've gotten rid of the message class entirely and folded it into chatops. Still need to figure out how we want to handle reactions. Can I get some thoughts on that ?

app/controllers/api/v1/chat_ops_controller.rb
```diff
@@ -17,7 +17,37 @@ def respond
...
+    user = User.find_by(slack_user_id: chat_params[:user_id])
+    return render_error("Slack `user_id` not found in retrodot", 403) unless user
+
+    result = ChatOps.process(user, chat_params[:text])
+
+    result = if result
+      result = result[:message] || result[:reaction]
+    else
+      "command unknown, try `/timeline help`"
+    end
```
reidmix: I'm thinking this could just be in `ChatOps#process`?  I'm not sure the pros/cons of having them together or separate, actually.

app/controllers/auth_controller.rb
```diff
@@ -56,14 +57,21 @@ def developer?
...
+    redirect_to admin_root_path, flash: { error: "Slack token already exists." }
   end
 
   protected
+
+  def update_slack_token
+    current_user.update(slack_access_token: slack_access_token)
+    log(fn: :install_slack, at: :update_slack_token, user: current_user.name)
+    redirect_to admin_root_path, notice: "Updated slack token"
+  end
```
reidmix: Could this be a part of `install_slack`?

reidmix: Should it be `User.with_slack_token.first.update(...)` so that we don't get multiple tokens?

reidmix: Dead Horse: I think Team is gonna make this less wonky.

joshuatobin: I'm going to do the `User.with_slack_token.first.update(...)` here and we'll address the team in a separate PR

app/controllers/api/v1/chat_ops_controller.rb
lexelby: One con is that the command (`"/timeline"`) is known here, whereas it's not known in `ChatOps.process` (unless we pass it in).  IMO this code should use `chat_params[:command]`, e.g. ```"command unknown, try `#{chat_params[:command]} help`"```.

config/routes.rb
```diff
@@ -22,6 +21,7 @@
     namespace :v1 do
       get  'chat_ops/matcher'
       post 'chat_ops/respond'
+      post "chat_ops/message"
```
lexelby: Since this method is Slack-specific, perhaps it should be `chat_ops/slack_message`? Or even `chat_ops/slack_slash_command`?

app/controllers/api/v1/chat_ops_controller.rb
reidmix: And TBH, I really do like that the presentation and the ChatOps logic is separate. 
I'm thinking leave it.

reidmix: you can get rid of the second `result`:
```ruby
result = if result
   result[:message] || result[:reaction]
else
  "command unknown, try `/timeline help`"
end
```

app/controllers/api/v1/chat_ops_controller.rb
```diff
@@ -17,7 +17,42 @@ def respond
...
   private
+
+  def format_result(result)
+    return default_message,   "ephemeral"   unless result
+    return result[:reaction], "ephemeral"   if result[:reaction]
+    return result[:message],  "in_channel"  if result[:message]
+  end
+
+  def default_message
+    "command unknown, try `#{chat_params[:command]} help`"
```
reidmix: later, but just a thought, later we can use the localization files to help store this copy.

joshuatobin: 👍

PR #82, opened 2016-10-24
Add migration for uniques (distincts)

joshuatobin:
@reidmix @lexelby can you give me some advice on how the best way would be to add a unique constraint on responders?

### DIFF
--- a/app/models/incident.rb
+++ b/app/models/incident.rb
@@ -3,7 +3,7 @@ class Incident < ActiveRecord::Base
   has_many :remediations, through: :retrospectives
   belongs_to :category
   has_many :timeline_entries, -> { order(timestamp: :asc) }
-  has_and_belongs_to_many :responders, join_table: :incidents_responders, class_name: "User"
+  has_and_belongs_to_many :responders, -> { distinct }, join_table: :incidents_responders, class_name: "User"
 
   default_scope { order('incident_id DESC') }
   scope :open, -> { where(state: "open") }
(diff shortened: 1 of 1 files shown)


reidmix:
Add a scope to the association `-> { uniq }` -- you may be able to remove the checks `unless responders...` 

https://stackoverflow.com/questions/16569994/deprecation-warning-when-using-has-many-through-uniq-in-rails-4

app/models/incident.rb
```diff
@@ -3,7 +3,7 @@ class Incident < ActiveRecord::Base
   has_many :remediations, through: :retrospectives
   belongs_to :category
   has_many :timeline_entries, -> { order(timestamp: :asc) }
-  has_and_belongs_to_many :responders, join_table: :incidents_responders, class_name: "User"
+  has_and_belongs_to_many :responders, -> { distinct }, join_table: :incidents_responders, class_name: "User"
```
joshuatobin: `uniq` changed in rails 5.1 -> https://github.com/rails/rails/commit/adfab2dcf4003ca564d78d4425566dd2d9cd8b4f

joshuatobin:
Verified adding myself multiple times: 
screen shot 2016-10-25 at 10 16 47 am

app/models/incident.rb
reidmix: oh interesting -- looks like they're trying to match sql syntax.

reidmix:
LGTM 👍

PR #83, opened 2016-10-25
Rake task for sync timeline

joshuatobin:
Simple rake task for syncing timeline. This will be removed post migration. 

CC: @reidmix @lexelby


reidmix:
👍

PR #88, opened 2016-11-01
Past incidents support

lexelby:
Add exporting of incident data in JSON format, and a mediator to sync our existing spreadsheet's extra columns into retrodot.

WIP for now -- just want a PR app.


lexelby:
@joshuatobin @reidmix This is ready for review (finally).  Here's a modified version of Past Incidents that works with this export code: https://docs.google.com/spreadsheets/d/1K6OxdE8liTlZt2kwNocmAe3l_Eg5cCACG9QSev6YV_A/edit#gid=0

I ended up modifying the existing code for the "automation -> update incidents" menu item rather than trying to use `IMPORTDATA()`. `IMPORTDATA()` doesn't allow me to specify formulas, and we have a lot of columns that depend on data in other columns.

app/support/mediators/incident/past_incidents_syncher.rb
```diff
@@ -0,0 +1,37 @@
...
+        log id: id
+
+        if id !~ /^\d+$/
+          log warning: "Skipping due to invalid incident ID", row: row.to_hash
+          next
+        end
+
+        category = ::Category.find_or_create_by(name: category_name) if category_name
+        team = ::Team.find_or_create_by(name: team_name)
+        incident = Incident.find_or_create_by(incident_id: id)
```
reidmix: Wow, I'd expect `Category` and `Team` _not_ to need the `::` anchor but `Incident` to need it because it would match `Mediators::Incident` -- I'm very confused.

app/support/mediators/incident/past_incidents_syncher.rb
```diff
@@ -0,0 +1,37 @@
...
+
+        log id: id
+
+        if id !~ /^\d+$/
+          log warning: "Skipping due to invalid incident ID", row: row.to_hash
+          next
+        end
+
+        category = ::Category.find_or_create_by(name: category_name) if category_name
+        team = ::Team.find_or_create_by(name: team_name)
```
reidmix: Would you need to guard against the `team_name` being `nil` as well?

app/models/team.rb
```diff
@@ -0,0 +1,2 @@
+class Team < ApplicationRecord
+end
```
reidmix: Should add a `team_name` unique validation.

spec/controllers/api/v1/export_controller_spec.rb
```diff
@@ -0,0 +1,10 @@
+RSpec.describe Api::V1::ExportController, type: :controller do
+
+  describe "GET #csv" do
+    it "returns http success" do
```
reidmix: I'd expect this spec to fail because it's not pointing at `export`

app/controllers/api/v1/export_controller.rb
```diff
@@ -0,0 +1,27 @@
+class Api::V1::ExportController < ApplicationController
+  skip_before_action :verify_authenticity_token
+  http_basic_authenticate_with name: "export", password: Config.export_api_key
+  respond_to :json
+
+  def export
+    limit = params.permit(:limit, :format)[:limit] || 100
```
lexelby: Actually, it _is_ being used, as far as I understand it, by rails.  I'm not sure if that means I need to permit it here, but the example I found said I did.

app/controllers/api/v1/export_controller.rb
```diff
@@ -0,0 +1,27 @@
...
+        category: incident.category&.name,
+        retro_at: incident.retro_at,
+        primary_team: incident.primary_team&.name,
+        followup: !!incident.followup_on,
+        trello_url: incident.trello_url,
+        google_doc_url: incident.google_doc_url
+      }
+    end
+
+    render json: output
```
lexelby: Is that preferable...?

app/support/mediators/incident/past_incidents_syncher.rb
lexelby: Oops, I sure do.  Thanks!

app/support/mediators/incident/past_incidents_syncher.rb
lexelby: ...me too!  I'll try it different ways.

app/controllers/api/v1/export_controller.rb
reidmix: I guess the `respond_to` is about returning the right error message (ie. not a missing template)?

I've used it in the past for the cleanup of the models `render` with `respond_with`, but I've not used both -- so I might be missing a nuance here?

app/controllers/api/v1/export_controller.rb
joshuatobin: hrm I don't believe you need the `:format` either.

joshuatobin:
Not a blocker but it might be helpful to have a readme on how this works in practice.

lexelby:
Honestly, our Past Incidents spreadsheet is going to go away pretty soon, so I don't really think we should document it.

app/controllers/api/v1/export_controller.rb
reidmix: i think its only doing the safe-guard in this method context.  I'm sure rails is using it too -- for sure -- but it wouldn't be using this instance of StrongParamaters AFAIK.

app/models/team.rb
reidmix: I meant validation to test uniqueness.

joshuatobin:
@lexelby LG :shipit:

PR #89, opened 2016-11-01
Return help message

joshuatobin:
Add logic for returning help message. 

CC: @lexelby @reidmix


reidmix:
Could be a Command and be treated like the other commands?

joshuatobin:
@reidmix re-jiggered it.

reidmix:
LGTM 👍

PR #90, opened 2016-11-02
Add formatting for slash commands

joshuatobin:
This cleans up formatting for the slack slash commands, and adds the ability to format certain messages with attachments. 

screen shot 2016-11-02 at 12 31 11 pm

CC: @lexelby @reidmix


lib/chat_ops/command.rb
```diff
@@ -60,7 +60,7 @@ def initialize(user, message)
     end
 
     def process
-      return unless @match = self.class.regex.match(@message)
+      return unless @match = self.class.regex.match(@message) rescue nil
```
joshuatobin: Without this an unknown command will throw:

```ruby
Failure/Error: return unless @match = self.class.regex.match(@message)

     NoMethodError:
       undefined method `match' for nil:NilClass
       Did you mean?  catch
```

app/controllers/api/v1/chat_ops_controller.rb
```diff
@@ -24,17 +24,29 @@ def slack_slash_command
...
+  def fmt_response(response_type: :ephemeral, attach_txt: default_message)
+    {
+      response_type: response_type,
+      text: chat_params[:text],
+      attachments: [{text: attach_txt, mrkdwn_in: [:text]}]
+    }
+  end
+
+  def help?
+    %w[help].include?(chat_params[:text])
```
reidmix: Or maybe `chat_params[:text] == "help"`?

app/controllers/api/v1/chat_ops_controller.rb
```diff
@@ -24,17 +24,29 @@ def slack_slash_command
...
 
   private
+  def fmt_result(result)
+    return fmt_response unless result
 
-  def format_result(result)
-    return default_message,   "ephemeral"   unless result
-    return result[:reaction], "ephemeral"   if result[:reaction]
-    return result[:message],  "in_channel"  if result[:message]
+    response_type = help?? :ephemeral : :in_channel
```
reidmix: I guess I'm confused now about the rules of response :D Could you help me by sharing the cases based on what the results are?

What I'm reading now it's saying that only help will be responded directly to the user (not in channel),  ALl other responses will be in channel and be attachments, where the contents of the result is the `:message` value (ie. no more `:reaction`).  Is that right?

If so, result doesn't need to be a hash anymore, it could just be the `:message` value.  

Or maybe the more general answer could be that the command could decide to have it be `:in_channel` or `:ephemeral`?

lib/chat_ops/commands/add_responder_command.rb
```diff
@@ -11,7 +11,7 @@ def run
 
       @incident.responders += responders
 
-      reaction(':checkmark:')
+      message("✅ got it!")
```
reidmix: seems like the response could be a `ChatOps::Command` constant or shared method (aka, `ok_response`)  since its used across many of the commands.  Or maybe use the localization features to store common text in `en.yml`

lib/chat_ops/commands/start_incident_command.rb
```diff
@@ -32,22 +32,20 @@ def run
...
       chat_start = get_chat_start(@match[:timestamp]).in_time_zone(Config.time_zone)
 
-      response = ["Recorded the start of chat for incident \##{incident_id} at #{chat_start.inspect}"]
+      response = ["Recorded the start of chat for incident `#{incident_id}` at `#{chat_start.inspect}`"]
 
       incident = Incident.find_or_create_by(incident_id: incident_id)
 
       if incident.chat_start
-        response << "     (overwriting start time for incident \##{incident_id}, was: #{incident.chat_start.inspect})"
+        response << "(*overwriting start time for incident* `#{incident_id}`, was: `#{incident.chat_start.inspect}`)"
```
reidmix: I'm not a huge fan of `inspect` in code outside of debugging, `inspect` is really `rfc822` although the format is slightly different.  Any way to change it?

app/controllers/api/v1/chat_ops_controller.rb
```diff
@@ -24,17 +24,29 @@ def slack_slash_command
...
 
     result = ChatOps.process(user, chat_params[:text])
-    msg, type = format_result(result)
 
-    render json: { text: msg, response_type: type }, status: 200
+    render json: fmt_result(result), status: 200
   end
 
   private
+  def fmt_result(result)
```
reidmix: Stylistic preference, but I like the full name `format_result` or `format` rather than abbreviations because it reads more naturally.

joshuatobin:
I'm completely re-working this so closing for a moment.

PR #91, opened 2016-11-03
remove test Command subclasses from ChatOps.command after each test

lexelby:
In `spec/lib/chat_ops/command_spec.rb`, we create subclasses of `Command`.  They get registered in `ChatOps.commands` and persist after the test completes, which can break other tests.  This PR saves `ChatOps.commands` and restores it after each test.

`spec/lib/chat_ops/chat_ops_spec.rb` already had code to do this, but it was buggy -- it should save a copy of `ChatOps.commands` instead of the actual Array object.

@joshuatobin  this fixes the bug you ran into: https://github.com/heroku/retrodot/pull/90#discussion_r86203836


joshuatobin:
👍

PR #92, opened 2016-11-04
Add support for starting incidents via webhooks

joshuatobin:
This PR adds the ability to start an incident via a webhook, and post the result to slack. The controller looks to see what type of `update_type` is passed, and if the incident update isn't `scheduled | maintenance`, and is the first update, then it will automagically run `start incident` and post the results to slack. 

CC: @lexelby @reidmix


app/controllers/api/v1/incidents_controller.rb
```diff
@@ -0,0 +1,57 @@
...
+  def post_to_slack
+    c = Slack::Client.new token: slack_token
+    c.chat_postMessage(
+      channel:  Config.slack_bot_channel,
+      username: Config.slack_bot_username,
+      text:     start_incident
+    )
+
+    true
+  end
```
reidmix: I might even decouple `start_incident` here to so that you could call it like: 
```ruby
post_to_slack(start_incident[:message])
```

app/models/incident.rb
```diff
@@ -30,6 +30,10 @@ def retro_prepared?
     trello_url && google_doc_url
   end
 
+  def missing_timeline?
+    timeline_entries.blank?
+  end
```
reidmix: <3

app/controllers/api/v1/incidents_controller.rb
```diff
@@ -0,0 +1,57 @@
...
+
+  def slack_token
+    User.with_slack_token.first.slack_access_token
+  end
+
+  def check_params
```
lexelby: What is this method for?  I don't see it used anywhere...

app/models/incident.rb
lexelby: I'd prefer `timeline_start.nil?` -- this would catch the case that someone has done `/t start incident` but not yet added anything to the timeline.

spec/controllers/api/v1/incidents_controller_spec.rb
```diff
@@ -0,0 +1,81 @@
+RSpec.describe Api::V1::IncidentsController do
+  let(:invalid_data) {{"incidents": 3124, "foo": "bar"}}
+  let!(:data)        {{"incident_id": 3227, "update_type"=>"issue"}}
```
lexelby: Can this rocket be a colon instead?

spec/controllers/api/v1/incidents_controller_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+      it "does not start incident" do
+        expect_any_instance_of(ChatOps::Commands::StartIncidentCommand).to_not receive(:process)
+
+        post :sync, params: data
+
+        expect(response).to have_http_status(:forbidden)
+      end
+
+      it "does not post to slack" do
+        expect_any_instance_of(Slack::Client).to_not receive(:chat_postMessage)
```
lexelby: Does this actually work?  I think I remember reading that this kind of test doesn't quite do what you want, since it will pass if _any_ instance of the class has not received the method call, even if some other instance _has_.  @reidmix do you know about this?

The way to test this would be to try putting this line in one of the tests that does post to slack and make sure it fails.

spec/controllers/api/v1/incidents_controller_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+
+  describe "Existing incidents" do
+    let!(:incident) { create(:incident, incident_id: 3227)}
+
+    it "starts an incident" do
+      expect_any_instance_of(ChatOps::Commands::StartIncidentCommand).to receive(:process).and_return({message: "foo"})
+
+      post :sync, params: data
+
+      expect(response).to have_http_status(:accepted)
```
lexelby: Rather than just trusting the response code, it would be a good idea here to verify that the timeline was started.  Perhaps test that `incident.timeline_started` is not nil?  Same for a similar test below.

app/models/incident.rb
reidmix: `nil?` won't work because it will return an empty array.  `empty?` will work too.

lexelby: Note, I said `timeline_start`, not `timeline_entries`.

reidmix: Ah, nice.

app/controllers/api/v1/incidents_controller.rb
```diff
@@ -0,0 +1,57 @@
...
+      render json: nil, status: 202
+    elsif !@incident
+      # If the incident does not exist, first sync before starting the incident.
+      Mediators::Incident::OneSyncher.run(id: params[:incident_id])
+      status = start_incident?? 202 : 403
+      render json: nil, status: status
+    else
+      render json: nil, status: 403
+    end
+  end
```
joshuatobin: @reidmix ah yeah. Way better thanks. I love the `can_start_incident?` ternary bit.

app/controllers/api/v1/incidents_controller.rb
```diff
@@ -0,0 +1,57 @@
...
+
+  private
+  def start_incident?
+    allowed_update_type?? post_to_slack : false
+  end
```
joshuatobin: Yeah +1

app/controllers/api/v1/incidents_controller.rb
joshuatobin: 👍

app/controllers/api/v1/incidents_controller.rb
joshuatobin: I just forgot to call it directly. Fixed.

app/controllers/api/v1/incidents_controller.rb
```diff
@@ -0,0 +1,57 @@
+require "slack"
+class Api::V1::IncidentsController < ApplicationController
+  skip_before_action :verify_authenticity_token
```
joshuatobin: 🌮

spec/controllers/api/v1/incidents_controller_spec.rb
joshuatobin: Yeah good catch. That was sloppy! thx

app/controllers/api/v1/incidents_controller.rb
```diff
@@ -0,0 +1,53 @@
...
+      render json: nil, status: 202
+    else
+      render json: nil, status: 403
+    end
+  end
+
+  private
+  def can_start_incident?
+    allowed_update_type? && @incident && @incident&.missing_timeline?? true : false
+  end
```
joshuatobin: @reidmix this is what I came up with ^

app/models/incident.rb
```diff
@@ -30,6 +30,10 @@ def retro_prepared?
     trello_url && google_doc_url
   end
 
+  def missing_timeline?
+    timeline_start.nil?
+  end
```
joshuatobin: @lexelby fixed here

joshuatobin:
@reidmix @lexelby Addressed all comments. Please take a look.

app/controllers/api/v1/incidents_controller.rb
reidmix: you'll get `false` if you have no incident (you want it to return `true`) -- that's why i ended up needing a ternary on whether or not there's an incident: `@incident ? @incident.missing_timeline? : true`

lexelby: @reidmix I think we don't actually need to return true if the incident doesn't exist.  He's having it sync the incident from Status if Retrodot doesn't know anything about it.  If Status is notifying us about an incident it knows nothing about, that's _really weird_ and we probably shouldn't just create one in retrodot.

app/controllers/api/v1/incidents_controller.rb
```diff
@@ -1,46 +1,42 @@
...
     else
       render json: nil, status: 403
     end
   end
 
   private
-  def start_incident?
-    allowed_update_type?? post_to_slack : false
+  def can_start_incident?
+    allowed_update_type? && @incident && @incident&.missing_timeline?? true : false
```
lexelby: Is the ternary doing anything here?  I think it could just be removed, and the last test can just be `@incident.missing_timeline?`

app/controllers/api/v1/incidents_controller.rb
```diff
@@ -0,0 +1,53 @@
...
+      render json: nil, status: 202
+    else
+      render json: nil, status: 403
+    end
+  end
+
+  private
+  def can_start_incident?
+    allowed_update_type? && @incident && @incident.missing_timeline?
+  end
```
joshuatobin: This should fix all of the concerns. We have to have a correct `update_type`, we have to have a valid incident, and one without a timeline.

app/controllers/api/v1/incidents_controller.rb
joshuatobin: I've made this much more simple by just removed the ternary.

spec/controllers/api/v1/incidents_controller_spec.rb
```diff
@@ -31,22 +32,20 @@
...
     describe "Existing incidents with a timeline" do
+      let!(:incident) { create(:incident, timeline_start: Time.now)}
       let!(:timeline) { create(:timeline_entry, incident: incident, user: user)}
 
       it "does not start incident" do
-        expect_any_instance_of(ChatOps::Commands::StartIncidentCommand).to_not receive(:process)
-
         post :sync, params: data
 
+        expect(incident.missing_timeline?).to be false
```
joshuatobin: Fixed

joshuatobin:
@lexelby Addressed all concerns.

app/controllers/api/v1/incidents_controller.rb
reidmix: I'm looking at the use-case where an incident isn't opened yet but we want to start a timeline.  I think before we would go ahead and create the timeline for an incident with an incremented ID -- I don't think we have that functionality anymore and not sure if that's intentional.   

When we retrieve no incident from the status site, it looks like it 500s for some reason, so no incident would be created.

That all being said, if we create or update an incident (even when there isn't one yet), the above logic would work

reidmix: OK, @joshuatobin explained that case is covered in the chapos `IncidentController` and that `@incident` would always be available in the Status webhook `IncidentController`.

PR #93, opened 2016-11-07
Add Formatting to commands

joshuatobin:
This PR adds the ability for individual commands to specify slack formatting. 

CC: @reidmix @lexelby


lib/chat_ops/commands/add_timeline_entry_command.rb
```diff
@@ -46,7 +46,7 @@ def run
 
       @incident.timeline_entries << ::TimelineEntry.new(user: @user, message: @match[:message])
 
-      reaction(':checkmark:')
+      message("✅ got it!")
```
reidmix: FYI, `Config.chatops_message_confirmation`

joshuatobin: Fixed!

app/controllers/api/v1/chat_ops_controller.rb
```diff
@@ -23,18 +23,32 @@ def slack_slash_command
     user = User.find_by(slack_user_id: chat_params[:user_id])
     return render_error("Slack `user_id` not found in retrodot", 403) unless user
 
-    result = ChatOps.process(user, chat_params[:text])
-    msg, type = format_result(result)
+    @result = ChatOps.process(user, chat_params[:text])
 
-    render json: { text: msg, response_type: type }, status: 200
+    render json: format_result, status: 200
```
lexelby: We really need to add a delay here.  Otherwise the fast response messes slack up and then everyone has to scroll down manually.  It'd be great if they fixed that bug, but until then, we have to around it. :(

app/models/incident.rb
```diff
@@ -23,7 +23,7 @@ def open?
   end
 
   def format_timeline
-    timeline_entries.map {|t| "#{t.timestamp.utc} #{t.user.name}: #{t.message}"}.join("\n")
+    timeline_entries.map {|t| "`#{t.timestamp.utc}` #{t.user.name}: #{t.message}"}.join("\n")
```
lexelby: Hmm.  This is starting to be pretty slack-specific, so it seems weird to me to have it in the model.  Maybe pull this out to TimelineCommand?

lib/chat_ops/commands/add_responder_command.rb
```diff
@@ -11,7 +11,7 @@ def run
 
       @incident.responders += responders
 
-      reaction(':checkmark:')
+      message(Config.chatops_message_confirmation)
```
lexelby: Can we write a helper function for this on `Command`?

lib/chat_ops/commands/timeline_command.rb
```diff
@@ -2,7 +2,7 @@ module ChatOps::Commands
   class TimelineCommand < ChatOps::Command
     setup do
       match /timeline(\s+(?<incident_id>\d+))?$/
-      help "timeline [#] - list the timeline for a incident"
+      help "`timeline [#]` - list the timeline for a incident"
```
lexelby: Can you also change the response below to be an attachment?

The current implementation formats it as an attachment and puts five newlines at the start.  That causes Slack to entirely collapse the timeline and just say "click for more...".  That way a long timeline doesn't fill everyone's screens.

lib/chat_ops/message_helpers.rb
```diff
@@ -8,16 +8,19 @@ def old_incident_warning
...
     end
 
-    def message(text)
-      { message: prevent_highlights(text) }
+    def message(text, response_type: nil, attachment: nil)
+      message = { message: prevent_highlights(text) }
+
+      # Give commands the ability to control their formatting individually
+      if response_type || attachment
+        message.update(response_type: response_type, attachment: attachment).compact
```
lexelby: What does the `compact` call do?  I don't think `Hash` has a `compact` method?

spec/helpers/chat_ops_command_helper.rb
```diff
@@ -6,7 +6,7 @@
 
 RSpec::Matchers.define :react_with do |expected|
   match do |actual|
-    actual && actual[:reaction] == expected
+    actual && actual[:message] == Config.chatops_message_confirmation
```
lexelby: This is kind of confusing, because the `expected` argument isn't even used anymore...

lib/chat_ops/message_helpers.rb
joshuatobin: Just removes the the nil. http://apidock.com/rails/Hash/compact

lexelby: Oh, of course, rails defines it. :facepalm: If you're compacting, do you need the `if`?

lib/chat_ops/commands/timeline_command.rb
joshuatobin: Fixed

lib/chat_ops/commands/add_responder_command.rb
joshuatobin: Yup. All set. Added `return_confirmation`

app/controllers/api/v1/chat_ops_controller.rb
joshuatobin: Added a 1/2 second delay

joshuatobin: nm fixing this in another PR

joshuatobin:
@lexelby @reidmix I've addressed all comments. Please take another look

lib/chat_ops/commands/timeline_command.rb
```diff
@@ -2,11 +2,13 @@ module ChatOps::Commands
...
     end
 
     def run
-      message(prevent_highlights("Timeline for incident #{@incident.incident_id}\n#{@incident.format_timeline}"))
+      timeline = @incident.timeline_entries.map {|t| "`#{t.timestamp.utc}` #{t.user.name}: #{t.message}"}.join("\n")
+
+      message("Timeline for incident `#{@incident.incident_id}`\n#{timeline}", attachment: true)
```
lexelby: Can you please make it like this:

`"Timeline for incident `#{@incident.incident_id}`\n\n\n\n\n#{timeline}"`

That will make it collapse by default.  You could even do this:

`"\n\n\n\n\nTimeline for incident `#{@incident.incident_id}`\n#{timeline}"`

joshuatobin: Ah I didn't understand what you meant before. Fixed.

joshuatobin:
@lexelby Addressed all comments.

lib/chat_ops/commands/add_responder_command.rb
```diff
@@ -11,7 +11,7 @@ def run
 
       @incident.responders += responders
 
-      reaction(':checkmark:')
+      return_confirmation
```
reidmix: 👍

lib/chat_ops/commands/timeline_command.rb
reidmix: maybe call the five newlines a constant that shows intent `FORCE_COLLAPSE = "\n"*5`

reidmix: could also be another formatting detail that could go into `format_response` because then the idiosyncrasies of Slack formatting ends up being localized in one method. 

Meaning, add an option: `collapse: true` and handle it later by adding the 5 newlines.

app/controllers/api/v1/chat_ops_controller.rb
```diff
@@ -1,6 +1,7 @@
 class Api::V1::ChatOpsController < ApplicationController
   skip_before_action :verify_authenticity_token
   http_basic_authenticate_with name: "api", password: Config.chatops_api_key, except: :slack_slash_command
+  FORCE_COLLAPSE = "\n" * 5
```
joshuatobin: @lexelby @reidmix Adding the constant here.

app/controllers/api/v1/chat_ops_controller.rb
```diff
@@ -23,18 +24,40 @@ def slack_slash_command
...
+  end
+
+  # All responses should default to in_channel without attachments.
+  def format_response(message:, response_type: :in_channel, attachment: nil, collapse: nil)
+    response = { response_type: response_type, text: message }
+
+    if attachment
+      attachment << FORCE_COLLAPSE if collapse
+      response[:attachments] = [{text: attachment, mrkdwn_in: [:text]}]
+    end
```
joshuatobin: @reidmix @lexelby Adding attachment collapse formatting here.

lib/chat_ops/commands/timeline_command.rb
joshuatobin: See comments above.

lib/chat_ops/message_helpers.rb
```diff
@@ -8,16 +8,16 @@ def old_incident_warning
       message "It looks like you may have forgotten to run `#{Config.chatops_prefix}start incident`.  If you really meant incident #{@incident.incident_id}, please specify the incident id with your command."
     end
 
-    def message(text)
-      { message: prevent_highlights(text) }
+    def message(text, **args)
+      { message: prevent_highlights(text), **args }.compact
+    end
```
joshuatobin: I like how this ended up turning out. Let me know if you see any problems with this approach.

joshuatobin:
@reidmix @lexelby Ready for review after adding collapse formatting option

app/controllers/api/v1/chat_ops_controller.rb
lexelby: The five newlines should be at the start of the text, not the end.

lib/chat_ops/message_helpers.rb
```diff
@@ -8,13 +8,8 @@ def old_incident_warning
...
 
-    def message(text, response_type: nil, attachment: nil)
-      message = { message: prevent_highlights(text) }
-
-      # Give commands the ability to control their formatting individually
-      message.update(response_type: response_type, attachment: attachment).compact
-
-      message
+    def message(text, **args)
+      { message: prevent_highlights(text), **args }.compact
```
lexelby: Whoa, neat trick!  I didn't know you could do that.

app/controllers/api/v1/chat_ops_controller.rb
reidmix: I'm not sure `collapse` is getting passed through.

reidmix: I do like that its here in this method tho 👍

reidmix: (Sorry I see the collapsed getting passed through now).

app/controllers/api/v1/chat_ops_controller.rb
```diff
@@ -23,18 +24,40 @@ def slack_slash_command
...
 
-  def format_result(result)
-    return default_message,   "ephemeral"   unless result
-    return result[:reaction], "ephemeral"   if result[:reaction]
-    return result[:message],  "in_channel"  if result[:message]
+    return format_response(
+      message:       chat_params[:text],
+      response_type: @result[:response_type] || :in_channel,
+      attachment:    @result[:message],
+      collapse:      @result[:collapse] || nil
```
reidmix: Don't need `|| nil`, if it's not in the `@result` it will already turn nil.  
This also has a side-effect of turning `false` into `nil`.
You could even force it into a boolean `!!@result[:collapse]`

lib/chat_ops/message_helpers.rb
reidmix: or `merge!`

joshuatobin:
@reidmix okay removed the `|| nil`.

PR #95, opened 2016-11-09
Add ability to add timeline events via slack reactions

joshuatobin:
This PR adds the ability to add a slack emoji reaction to a message and have that message added to the timeline of the currently running incident. It uses the slack events API to subscribe to all `reaction_added` events, and filters the reaction out based upon a configureable reaction name. The endpoint will only add reactions to incidents that are currently `in_progress?`.  

Reasons for doing this: 
* Spent most of my weekend in airports / on a plane.
* I seem to spend a majority of my time in an incident copying/pasting things others have noted back into the timeline. 
* Seemed like it would be fun.  

CC: @lexelby @reidmix


app/controllers/api/v1/chat_ops_controller.rb
```diff
@@ -57,6 +77,35 @@ def format_response(message:, response_type: :in_channel, attachment: nil, colla
...
+
+  # Required by slack to setup a new events endpoint
+  def render_slack_challenge
+    render json: { challenge: event_params[:challenge] }, status: 200
+  end
+
+  def slack_reaction?
+    ["reaction_added"].include?(event_params[:event][:type]) &&
+    ["message"].include?(event_params[:event][:item][:type]) &&
+    [Config.timeline_reaction_emoji].include?(event_params[:event][:reaction])
```
lexelby: Why array-ify and use `include?`? Wouldn't `==` work fine?

joshuatobin: Yeah either way works.

joshuatobin: fixed

app/controllers/api/v1/chat_ops_controller.rb
```diff
@@ -57,6 +77,35 @@ def format_response(message:, response_type: :in_channel, attachment: nil, colla
...
+      latest:    event_params[:event][:item][:ts],
+      oldest:    event_params[:event][:item][:ts],
+      inclusive: 1
+    )["messages"].first["text"]
+  end
+
+  # Required by slack to setup a new events endpoint
+  def render_slack_challenge
+    render json: { challenge: event_params[:challenge] }, status: 200
+  end
```
reidmix: So we just echo the `challenge` param back to the requester?

app/controllers/api/v1/chat_ops_controller.rb
```diff
@@ -19,16 +20,35 @@ def respond
...
+  def slack_event
+    return render_slack_error(t('slack.errors.token'), 401) unless valid_slack_token?
+    return render_slack_challenge if slack_url_verification?
+
+    # Dont process events unless we're in the midst of an incident
+    unless Incident.in_progress?
+      message = t('slack.errors.incident.missing')
+      post_to_slack(format_response(message: message, response_type: :ephemeral), slack_channel)
+      return render json: message, status: 403
+    end
```
reidmix: Would this show the error every time anyone uses any emoji when there's no incident?

I wonder if we'd be spamming slack users with error messages on every emoji... I'm betting I'm missing something.

app/controllers/api/v1/chat_ops_controller.rb
```diff
@@ -19,16 +20,35 @@ def respond
...
+
+    if slack_reaction?
+      post_to_slack(add_timeline_entry(slack_message_from_timestamp, Incident.open_incident_id)[:message], slack_channel)
```
reidmix: I think for me, personally, it would be easier to read to break it down into parts because there is a lot going on:
```ruby
result = add_timeline_entry(slack_message_from_timestamp, Incident.open_incident_id)
post_to_slack(result[:message], slack_channel)
```

reidmix: Does this happen on every channel when there's an incident and someone uses that `timeline_reaction_emoji`?  

I'm wondering if it should be constrained because what if random messages get tagged because that emoji just happened to be used.

app/controllers/api/v1/chat_ops_controller.rb
```diff
@@ -57,6 +77,35 @@ def format_response(message:, response_type: :in_channel, attachment: nil, colla
...
+  end
+
+  def slack_message_from_timestamp
+    c = Slack::Client.new token: slack_token
+    c.channels_history(
+      channel:   event_params[:event][:item][:channel],
+      latest:    event_params[:event][:item][:ts],
+      oldest:    event_params[:event][:item][:ts],
+      inclusive: 1
+    )["messages"].first["text"]
```
reidmix: what if something goes wrong for some reason getting this history back?

app/controllers/api/v1/chat_ops_controller.rb
```diff
@@ -65,16 +114,22 @@ def default_message
...
+    )
+  end
+
+  def get_slack_user
+    User.find_by(slack_user_id: event_params[:event][:user])
+  end
+
+  def slack_channel
+    event_params[:event][:item][:channel]
   end
```
reidmix: all of these deep hashes kind suck (ie. `event_params[:event][:item][:channel]`).  I like what you did here to name them via a method I wonder if you should be doing it more?  (ie. `slack_user`, `slack_channel`, `slack_reaction`, etc.)?  I'm not sure if that's better or more obfuscated?

app/controllers/api/v1/chat_ops_controller.rb
joshuatobin: Yeah this is a one time thing, but necessary to properly setup the endpoint on slacks side. Once you "echo" it back to slack, they'll start forwarding events on. More here: https://api.slack.com/events-api#subscriptions

app/controllers/api/v1/chat_ops_controller.rb
joshuatobin: Yeah I went back and forth on that. I'm happy to do so but thought it might be overkill. I broke the specific ones out that I thought might be re-usable.

app/controllers/api/v1/chat_ops_controller.rb
reidmix: Ah thank you.

app/controllers/api/v1/chat_ops_controller.rb
reidmix: I think that's ok for now. 👍

app/controllers/api/v1/chat_ops_controller.rb
joshuatobin: Sure I"ll break it out, and I agree i was trying to do to much.

joshuatobin: Good point. Yeah'll I'll also work on retraining it to one configurable channel.

app/controllers/api/v1/chat_ops_controller.rb
joshuatobin: Yup needs some error handling. Do you think its best to add a rescue or just a `try`?

reidmix: I think a `rescue` because you could have
- `c.channels_history` coming back with an error (or bad result)
- and a bad result where the `first` wouldn't work, in that case a `try` would give you `nil` and you should basically end stop there.

app/controllers/api/v1/chat_ops_controller.rb
```diff
@@ -19,16 +20,35 @@ def respond
...
 
+  def slack_event
+    return render_slack_error(t('slack.errors.token'), 401) unless valid_slack_token?
+    return render_slack_challenge if slack_url_verification?
+
+    # Dont process events unless we're in the midst of an incident
+    unless Incident.in_progress?
+      message = t('slack.errors.incident.missing')
+      post_to_slack(format_response(message: message, response_type: :ephemeral), slack_channel)
```
joshuatobin: @reidmix maybe I could just tack on `if slack_reaction?` at the end here. That would limit it to our custom emoji

app/controllers/api/v1/chat_ops_controller.rb
joshuatobin: @reidmix Also thinking more about this, it might be to our advantage to be able to react from different channels.

app/controllers/api/v1/chat_ops_controller.rb
joshuatobin: Thx I'll do that.

app/controllers/api/v1/chat_ops_controller.rb
joshuatobin: @reidmix Added some error handling in bdaa3bc

joshuatobin:
@reidmix I've addressed all concerns except being able to "react" from different channels. I'd personally like to be able to add a timeline reaction from other than just one channel, but I can see pros and cons for this. Otherwise this should be ready for review.

reidmix:
I feel better that react is based on a custom emoji (a special emoji 😄) and we will then be able to see if rando messages get tied accidentally to a timeline.

app/controllers/api/v1/chat_ops_controller.rb
```diff
@@ -57,6 +80,37 @@ def format_response(message:, response_type: :in_channel, attachment: nil, colla
...
+
+  def slack_message_from_timestamp
+    c = Slack::Client.new token: slack_token
+    c.channels_history(
+      channel:   event_params[:event][:item][:channel],
+      latest:    event_params[:event][:item][:ts],
+      oldest:    event_params[:event][:item][:ts],
+      inclusive: 1
+    )["messages"].first["text"]
+  rescue Slack::Error, NoMethodError
```
reidmix: I'd say catch any error here

app/controllers/api/v1/chat_ops_controller.rb
```diff
@@ -57,6 +80,37 @@ def format_response(message:, response_type: :in_channel, attachment: nil, colla
...
+    raise NoMethodError, "slack message not found"
```
reidmix: Raise a specialized error (ie. define our own error). or a `Slack::Error` here.
Not throwing a `RuntimeError` seems right to me, which I think is what your intention was.

app/controllers/api/v1/chat_ops_controller.rb
```diff
@@ -19,16 +20,38 @@ def respond
...
+      message = t('slack.errors.incident.missing')
+      post_to_slack(format_response(message: message, response_type: :ephemeral), slack_channel)
+      return render json: message, status: 403
+    end
+
+    entry = add_timeline_entry(slack_message_from_timestamp, Incident.open_incident_id)
+    post_to_slack(entry[:message], slack_channel)
+
+    render json: t('slack.timeline.added'), status: 201
+  rescue Slack::Error, NoMethodError => e
```
reidmix: Log the error too, please.

reidmix: We can remove `NoMethodError` here (see my comment on not raising a `NoMethodError`)

app/controllers/api/v1/chat_ops_controller.rb
reidmix: My thinking is that `NoMethodError` isn't semantically right here to re-raise.

app/controllers/api/v1/chat_ops_controller.rb
```diff
@@ -33,6 +33,7 @@ def slack_slash_command
   def slack_event
     return render_slack_error(t('slack.errors.token'), 401) unless valid_slack_token?
     return render_slack_challenge if slack_url_verification?
+    return render json: t('slack.errors.reaction'), status: 403 unless slack_reaction?
```
lexelby: I'm not sure what this is supposed to do.  As far as I can see, `t('slack.errors.reaction')` is going to be the string `"Invalid timeline reaction"`, so not JSON data at all.  Also, we don't want the bot to complain on the channel every time someone reacts to a message using a reaction that isn't the special timeline reaction.

app/controllers/api/v1/chat_ops_controller.rb
joshuatobin: @reidmix Fixed the exception handling. Can you check again?

joshuatobin:
Updated with latest rubocop fixes. CC: @reidmix @ys @stellacotton @atmos

app/controllers/api/v1/chat_ops_controller.rb
```diff
@@ -20,16 +21,40 @@ def respond
...
+      post_to_slack(format_response(message: message, response_type: :ephemeral), slack_channel)
+      return render json: message, status: 403
+    end
+
+    entry = add_timeline_entry(slack_message_from_timestamp, Incident.open_incident_id)
+    post_to_slack(entry[:message], slack_channel)
+
+    render json: t("slack.timeline.added"), status: 201
+  rescue Slack::Error
+    error = "#{t('slack.errors.timeline')}: #{$ERROR_INFO}"
```
reidmix: Is rubocop making you use the long-form name instead of `$!` ?
(not sure I like that one)

reidmix:
👍

app/controllers/api/v1/chat_ops_controller.rb
joshuatobin: @reidmix Yeah thats a rubocop thing. I'm still not sold on that default but its something we can discuss.

joshuatobin: @atmos okay cool. I'll cue up a change for that. thanks

PR #97, opened 2016-11-10
Redirect bug fix

joshuatobin:
The `return_to` path for `GoogleAuthRequired` had a bug where it was set to the resource's id, vs the path + resource id. This was causing the callback to blow up.

CC: @lexelby @reidmix


joshuatobin:
Simple fix so merging this.

app/admin/incident.rb
```diff
@@ -37,7 +37,7 @@ class TrelloAuthRequired < StandardError; end
     redirect_to resource_path, notice: "Synced!"
   end
 
-  member_action :prepare_retro, method: :post do
+  member_action :prepare_retro, method: [:post, :get] do
```
joshuatobin: @reidmix I actually had to add the `GET` because it was barfing when it was getting redirected with `No route found`

joshuatobin: Oh sorry. Its late :) I'll triple check and get back on this.

joshuatobin: @reidmix - So the `get` will flash a message upon state change. Maybe we can pair on this if there are other actions we can do to prevent accidental changes?

reidmix: Yes, let's do that :D

PR #99, opened 2016-11-11
Add cron along with error handling for syncing slack tokens

joshuatobin:
CC: @reidmix @lexelby


app/support/mediators/user/sync_slack_users.rb
```diff
@@ -4,12 +4,13 @@ module Mediators::User
   class SyncSlackUsers < Mediators::Base
 
     def initialize(token:)
+      abort("Invalid token") if token.nil?
```
reidmix: Maybe "Missing token" as an error message?

joshuatobin:
All set.

PR #101, opened 2016-11-15
Add links for trello/google doc to the show view

joshuatobin:
I've been trying to find a better way to do this, but have not yet succeeded so this is a HACK. ActiveAdmin allows for customization of the show view by allowing access to the `attributes_table`. This unfortunately allows you to only display specific attributes at a time. I was however able to find the `default_attribute_table_rows` method, which returns an array of symbols of the field names available for a resource, however I'm still having to convert the symbol to a string before doing the evaluation.

What I'd like to do is something more along the lines of: 

```ruby
     default_attribute_table_rows.each do |field, value|
        if value =~ /^http:?\/\//
          row "#{link_to(field, value)}
        else
          row field
        end
```

CC: @reidmix maybe you have a better idea of how to do this?


app/admin/incident.rb
```diff
@@ -7,6 +7,20 @@ class TrelloAuthRequired < StandardError; end
 ActiveAdmin.register Incident do
   config.sort_order = 'incident_id_desc'
 
+  show do
+    attributes_table do
```
joshuatobin: Great thanks! Yeah would <3 to have that on by default

PR #102, opened 2016-11-15
Add confirmation to admin UI for email delivery

joshuatobin:
Adds some visual confirmation to the UI if an email was actually sent out, and removes the 'Sent Retro Email' button if so.

screen shot 2016-11-15 at 1 03 21 pm

CC: @reidmix


app/admin/incident.rb
```diff
@@ -126,7 +134,7 @@ class TrelloAuthRequired < StandardError; end
           { To:      :text,
             CC:      :text,
             Subject: :text
-          }.to_json
+          }.to_json unless incident.retro_email_sent?
```
ys: I find it hard to see what is the scope of the unless here... with the two levels of hashes and different syntax.

db/schema.rb
```diff
@@ -53,6 +53,7 @@
     t.string   "google_doc_url"
     t.datetime "retro_at"
     t.integer  "primary_team_id"
+    t.boolean  "retro_email_sent"
```
ys: Please add a default to false here. You have a three legs boolean.

db/schema.rb
```diff
@@ -42,8 +42,8 @@
     t.integer  "incident_id"
     t.boolean  "review"
     t.date     "followup_on"
-    t.datetime "created_at",      null: false
-    t.datetime "updated_at",      null: false
+    t.datetime "created_at",       null: false
```
joshuatobin: @ys Not sure. This is automated AFAICT

db/schema.rb
joshuatobin: Fixed in ae30557 @ys

app/admin/incident.rb
joshuatobin: Not sure it reads as well but fixed in 0c6f277

joshuatobin:
@ys Addressed all comments.

PR #104, opened 2016-11-16
Redirect root to /admin

joshuatobin:
Redirect `/` -> `/admin` until we actually have a useful index page. 

CC: @reidmix @ys


config/routes.rb
```diff
@@ -1,6 +1,6 @@
 Rails.application.routes.draw do
+  root to: redirect("/admin")
```
ys: Why not make `/` be `/admin` ?

joshuatobin: @ys good call. fixed.

joshuatobin:
For now just activeadmin, but its just a stop gap for the MVP

PR #105, opened 2016-11-16
Add Rubocop

ys:
This adds rubocop with what we use in all our projects.
This is a PR open to discussion so that we can see what fits the best here.

Rubocop will allow to not have discussions around code style, avoid bikeshedding and so on:)


joshuatobin:
👍

PR #106, opened 2016-11-16
Rubocop Autocorrect almost all the things

ys:
I skipped the Regexes and the command_spec which was breaking for I don't know yet which reason.


app/admin/incident.rb
```diff
@@ -66,27 +66,27 @@ def prepare_document_links
         current_user: current_user
       )
     rescue TrelloAuthRequired
-      log_error($!, at: :create_trello_card, fn: :member_action)
+      log_error($ERROR_INFO, at: :create_trello_card, fn: :member_action)
```
joshuatobin: Didn't know this existed. Not sure the change is necessary but either way. https://ruby-doc.org/stdlib-2.0.0/libdoc/English/rdoc/English.html

app/models/user.rb
```diff
@@ -7,7 +7,7 @@ class User < ApplicationRecord
   scope :with_name,        -> { where.not(name: nil) }
   scope :with_slack_token, -> { where.not(slack_access_token: nil) }
 
-  def as_json(options={})
+  def as_json(_options = {})
```
joshuatobin: @ys why is this getting renamed?

app/admin/incident.rb
ys: This is less obscure. more explicit

app/models/user.rb
amerine: It looks like it's never used and hitting the UnusedMethodArgument rubocop linter.

lib/chat_ops/commands/end_incident_command.rb
```diff
@@ -9,7 +9,7 @@ def run
       @incident.chat_end = Time.now
       @incident.save
 
-      return { message: "Recorded the end of chat for incident `#{@incident.incident_id}` at `#{@incident.chat_end.inspect}`." }
+      { message: "Recorded the end of chat for incident `#{@incident.incident_id}` at `#{@incident.chat_end.inspect}`." }
```
joshuatobin: nice.

lib/chat_ops/command.rb
```diff
@@ -65,9 +66,9 @@ def process
       if @match.names.include? "incident_id"
         @incident = determine_incident(@match[:incident_id])
 
-        if !self.class.incident_optional?
-          return unknown_incident_warning if !@incident
-          return old_incident_warning if @match[:incident_id].nil? and @incident.old?
+        unless self.class.incident_optional?
+          return unknown_incident_warning unless @incident
+          return old_incident_warning if @match[:incident_id].nil? && @incident.old?
```
ys: Using `and` can have strange side effects if I recall correctly

spec/lib/chat_ops/chat_ops_spec.rb
```diff
@@ -55,28 +55,36 @@
...
+      cmd_1_class.class_eval do
+        def run
+          raise Exception("cmd_1_class instance's run() method should not be called")
+                               end
+      end
+      cmd_2_class.class_eval do
+        def run
+          "Command2.run"
+                               end
+      end
```
joshuatobin: This is kinda hard to read.

lib/chat_ops/command.rb
joshuatobin: 👍

joshuatobin:
@ys Everything looks good except my last comment in the `chat_ops_spec`.

joshuatobin:
👍

spec/lib/chat_ops/chat_ops_spec.rb
ys: this is not ok:D for sure:D

joshuatobin:
👍 🚢 :shipit:

Gemfile
```diff
@@ -1,64 +1,64 @@
...
+  gem "webmock"
+  gem "timecop"
 end
 
 group :development do
-  gem 'spring'
-  gem 'spring-watcher-listen', '~> 2.0.0'
+  gem "spring"
+  gem "spring-watcher-listen", "~> 2.0.0"
 end
```
reidmix: hooray for double quotes!

app/controllers/api/v1/chat_ops_controller.rb
```diff
@@ -66,7 +67,9 @@ def default_message
   end
 
   def valid_slack_token?
-    ActiveSupport::SecurityUtils.secure_compare(chat_params[:token], Config.slack_slash_command_token) rescue false
+    ActiveSupport::SecurityUtils.secure_compare(chat_params[:token], Config.slack_slash_command_token)
+  rescue
+    false
```
reidmix: Nice, I like this.  @ys do you know the advantage to using the method-level `rescue` here?  (Clearly if we had two lines of code, one with a rescue, then this wouldn't be the case).

app/models/user.rb
reidmix: Would it allow `as_json(_ = {})`?

app/support/mediators/incident/past_incidents_syncher.rb
```diff
@@ -14,7 +14,11 @@ def call
         id = row[ID_COLUMN]
         category_name = row[CATEGORY_COLUMN]
         team_name = row[TEAM_COLUMN]
-        retro_date = DateTime.strptime(row[RETRO_DATE_COLUMN], '%m/%d/%Y') rescue nil
+        retro_date = begin
+                       DateTime.strptime(row[RETRO_DATE_COLUMN], "%m/%d/%Y")
+                     rescue
+                       nil
+                     end
```
reidmix: This seems overly specified to me. Can you explain why wouldn't the `rescue` modifier be sufficient? (it's also nice and terse).

reidmix:
That wasn't really too terrible. Nicely done.

PR #107, opened 2016-11-16
Rubocop fixes for lib

joshuatobin:
CC: @reidmix @ys @stellacotton @atmos


.rubocop.yml
```diff
@@ -57,7 +57,7 @@ PercentLiteralDelimiters:
     '%w': '{}'
 
 LineLength:
-  Max: 80
+  Max: 150
```
atmos: You'll probably want to go back and revert this after everything gets to a passing point. You can disable the long lines on a case by case basis when absolutely necessary.

atmos:
:+1:

PR #108, opened 2016-11-16
Get rid of .env.test.sample 

joshuatobin:
.env.test.sample -> .env.test

CC: @reidmix @ys


.travis.yml
```diff
@@ -4,7 +4,7 @@ rvm:
 sudo: false
 bundler_args: --without development
 before_script:
-  - cp .env.test{.sample,} && cp .env.development{.sample,}
+  - cp .env.{test,} && cp .env.development{.sample,}
```
reidmix: I'd say just leave it in place -- but this works too.

PR #109, opened 2016-11-16
Fixes for rubocop in mediators

joshuatobin:
CC: @ys @reidmix


app/support/mediators/timeline/sync_timeline.rb
```diff
@@ -38,6 +40,7 @@ def data
       MultiJson.decode(get_timeline.body)
     end
 
+    # rubocop:disable AccessorMethodName
```
atmos: We generally fix these rather than disable them.

PR #110, opened 2016-11-17
Use Time.now.utc everywhere where it was Time.now

ys:
This forces the timezone to be the same everywhere we store data with `Time.now`.


joshuatobin:
👍 thanks @ys

joshuatobin:
👍

PR #121, opened 2016-11-30
Add logic to refresh google auth token

joshuatobin:
CC: @reidmix @jmervine


app/controllers/auth_controller.rb
```diff
@@ -74,13 +78,49 @@ def login_and_create_user
     flash[:notice] = "Logged in."
   end
 
+  def google_request_token
+    url = URI("https://accounts.google.com/o/oauth2/token")
+    Net::HTTP.post_form(url, token_refresh_params)
```
jmervine: Should this be switched to Excon for release, or are we good with sticking with Net::HTTP?

app/controllers/auth_controller.rb
```diff
@@ -74,13 +78,49 @@ def login_and_create_user
...
+      client_secret: Config.google_client_secret,
+      grant_type:    :refresh_token
+    }
+  end
+
   def update_google_creds
     creds = omniauth_info["credentials"]
-    current_user.update(google_refresh_token: creds[:refresh_token]) unless creds[:refresh_token].nil?
-
+    update_refresh_token unless creds[:refresh_token].nil?
```
jmervine: Might want to log on nil refresh_token, that shouldn't really happen, right?

db/migrate/20161130185605_add_refresh_token_expires_at.rb
```diff
@@ -0,0 +1,5 @@
+class AddRefreshTokenExpiresAt < ActiveRecord::Migration[5.0]
+  def change
+    add_column :users, :refresh_token_expires_at, :integer
```
jmervine: Why `:integer` and not `:datetime` (or whatever it is)?

app/controllers/auth_controller.rb
joshuatobin: No, this will happen a bunch. If a user logs out, but the token is still valid on googles end, it won't send back the refresh token.

db/migrate/20161130185605_add_refresh_token_expires_at.rb
joshuatobin: Good call.

jmervine: This change may also change the way you test expiry, removing the need for `to_i`.

jmervine:
👍

joshuatobin:
@jmervine - Can you review again. Tested this with an expired token after latest round of fixes and it works as expected

app/controllers/auth_controller.rb
```diff
@@ -74,13 +77,48 @@ def login_and_create_user
...
+
+  def refresh_token!
+    response = google_request_token
+    data = JSON.parse(response.body)
+    current_user.update(
+      google_refresh_token:     data["access_token"],
+      refresh_token_expires_at: Time.now.utc + data["expires_in"].to_i.seconds
+    )
+    log(fn: :refresh_token!, at: :update, expires_at: current_user.refresh_token_expires_at)
+  end
```
reidmix: I'd say log any error condition here and re-raise (or maybe do it on the calling method)

app/controllers/auth_controller.rb
```diff
@@ -74,13 +77,50 @@ def login_and_create_user
     flash[:notice] = "Logged in."
   end
 
-  def update_google_creds
+  def google_request_token
+    Excon.post("https://accounts.google.com/o/oauth2/token", token_refresh_params)
+  rescue Excon::Error
+    log_error
```
joshuatobin: @reidmix catching and logging here

joshuatobin:
@reidmix @jmervine can one of you take another look at this?

jmervine:
👍

PR #122, opened 2016-12-01
Fix document links

joshuatobin:
Open the document links in a new tab.

screen shot 2016-12-01 at 11 51 19 am

screen shot 2016-12-01 at 11 52 05 am

CC: @jmervine


reidmix:
👍

PR #123, opened 2016-12-01
Add responders regex fix'n

joshuatobin:
Per user testing feedback, adding `to incident` to the chatops command for adding responders was redundant. Removing this and making sure we can add multiple users separated either by spaces or commas. Credit to @jmervine for the actual hard work done here 💃 

CC: @reidmix @jmervine


jmervine:
👍

PR #124, opened 2016-12-01
Update remove responders regex

joshuatobin:
CC: @reidmix @jmervine


lib/chat_ops/commands/remove_responder_command.rb
```diff
@@ -2,8 +2,8 @@ module ChatOps::Commands
   # Remove responder from an incident
   class RemoveResponderCommand < ChatOps::Command
     setup do
-      match(/remove\s+(?<who>.*)\s+from\s+incident(\s+(?<incident_id>\d+))?/)
-      help "`remove <person> [<person>...] [from <incident>]` -- <person> can be a full name or @handle"
+      match(/remove\s+(?<who>((?!\s\d+).)*)\s*((?<incident_id>\d+)?)\s*$/)
```
jmervine: Perhaps look at doing something like...

```ruby
match_users_and_incident_with("remove")

# where method like
def match_users_and_incident_with cmd
  match(/#{cmd}\s+(?<who>((?!\s\d+).)*)\s*((?<incident_id>\d+)?)\s*$/)
end
```

jmervine:
👍

PR #125, opened 2016-12-02
removing assets from procfile, done automatically

jmervine:
release runs `assets:precompile`, don't need it in the procfile, just slows things down

/cc @reidmix @joshuatobin


joshuatobin:
👍

PR #126, opened 2016-12-02
Remove sync button

joshuatobin:
CC: @jmervine @reidmix


jmervine:
👍

lexelby:
Why did we remove the sync button?

joshuatobin:
Its run out of cron on the platform. We didn't find a  need for it in the UI.

lexelby:
Oh! Now I remember.  We added that button before we had the cron, just for testing.

PR #127, opened 2016-12-02
adding error handling for missing refresh_token

jmervine:
/cc @reidmix @joshuatobin 

This is stupid. What we should really do is just not handle this **super**-edge case error condition where someone goes to retrodot, having authorized the app previously, but before the refresh_token handling was added today (https://github.com/heroku/retrodot/pull/121).

Instead, we should just tell everyone that's touched it to go to https://security.google.com/settings/security/permissions, deauth the app, logout, wait 5 minutes and then move forward.

Mainly posting PR this b/c we spent a bunch of time trying to find an elegant solution, ended with this ugly solution and wanted to open a dialog on it before scrapping it.

All that said, I 👎 this PR. =D


reidmix:
Have you talked with the Tools team who may have encountered this scenario?

jmervine:
@reidmix Haven't, but that's not a bad idea.

joshuatobin:
This is only a side effect that would have happened previously in development. Lets close this out for now as it shouldn't be an issue in production.

PR #128, opened 2016-12-03
command overhaul

jmervine:
/cc @joshuatobin @reidmix 

Based on conversation regarding updating commands to match the help message proposed in [this gist](https://gist.github.com/jmervine/2a52636f15dfc8c1f3c229d1f487c387).

- Simplified timeline commands
- Updated timeline help messages
- Added sensible shortcuts (undocumented in help)
- Refactored class names to match updated commands
- Merged `timeline` command actions in to a single regex / class

Note:
- Spec refactors are to support changes and instabilities.


spec/lib/chat_ops/commands/start_incident_command_spec.rb
```diff
@@ -3,17 +3,17 @@
...
+      start at 09:30
+      start 5 minutes ago
+      start 300 5 minutes ago
+      s 300 5 minutes ago
     EOL
   end
 
   describe ".run" do
-    command :start_incident
+    command :start
```
jmervine: I recommend removing the `command` spec helper all together, as it's only used in a single spec and a bit obfuscated.

spec/lib/chat_ops/commands/add_timeline_entry_command_spec.rb
```diff
@@ -8,10 +8,12 @@
       timeline 12 foo
       timeline 12 foo bar baz
       timeline 1is the loneliest number
+      t foo
     EOL
 
     it_should_not_match_commands <<-EOL
       timeline 12
+      t 12
```
joshuatobin: <3

spec/lib/chat_ops/commands/start_incident_command_spec.rb
joshuatobin: Works for me.

joshuatobin:
@jmervine Great job. I really like the shortcuts (`rm`) etc that you've done for the commands.

joshuatobin:
👍

lib/chat_ops/commands/add_command.rb
```diff
@@ -0,0 +1,17 @@
+module ChatOps::Commands
+  # Add responders to timeline
+  class AddCommand < ChatOps::Command
+    setup do
+      match(/(add|a|\+)($|\s+)((?<incident_id>\d+)?)\s*(?<who>((?!\s\d+).)*)\s*$/)
```
reidmix: That's nice (the solo `add` and the negative-look ahead)

lib/chat_ops/commands/end_command.rb
```diff
@@ -1,9 +1,9 @@
 module ChatOps::Commands
   # End the incident
-  class EndIncidentCommand < ChatOps::Command
+  class EndCommand < ChatOps::Command
     setup do
-      match(/end\s+(the\s+)?incident(\s+(?<incident_id>[0-9]+))?/)
-      help "`end incident [#]` - sets the end of chat for an incident"
+      match(/(end|e|stop)($|\s+)(\s+(?<incident_id>[0-9]+))?/)
```
reidmix: personally, I don't think `/t e` is that useful.

lib/chat_ops/commands/add_command.rb
reidmix: personally, I don't think `/t a` is that useful.

lib/chat_ops/commands/responders_command.rb
```diff
@@ -1,9 +1,9 @@
 module ChatOps::Commands
   # List responders for an incident
-  class ListResponderCommand < ChatOps::Command
+  class RespondersCommand < ChatOps::Command
     setup do
-      match(/incident(\s+(?<incident_id>\d+))?\s+responders/)
-      help "`incident [#] responders` - list responders for an incident"
+      match(/(responders|r|who)(\s+(?<incident_id>\d+))?\s*/)
```
reidmix: personally, I don't think `/t r` is that useful.  but loving `/t who`

lib/chat_ops/commands/timeline_command.rb
```diff
@@ -1,12 +1,32 @@
 module ChatOps::Commands
-  # Lists the timeline for an incident
+  # Add timeline entries
   class TimelineCommand < ChatOps::Command
     setup do
-      match(/timeline(\s+(?<incident_id>\d+))?$/)
-      help "`timeline [#]` - list the timeline for a incident"
+      match(/(timeline|t)\s*(?<incident_id>\d+)?\s*(?<message>.+)?\s*$/)
```
reidmix: how about `t log` as a shortcut?

lib/chat_ops/commands/timeline_command.rb
```diff
@@ -1,12 +1,32 @@
...
+
+      help "`timeline [#] <message>` - add message to incident timeline\n`timeline [#]` - list details of an incident timeline"
```
reidmix: Maybe for code readability you could:
```ruby
help [
  "`timeline [#] <message>` - add message to incident timeline",
  "`timeline [#]` - list details of an incident timeline"
].join("\n")
```

Or alter `help` so that:
```ruby
help \
  "`timeline [#] <message>` - add message to incident timeline",
  "`timeline [#]` - list details of an incident timeline"
```

lib/chat_ops/commands/timeline_command.rb
```diff
@@ -1,12 +1,32 @@
...
     end
 
     def run
+      @match[:message] ? add_timeline_message : show_timeline
```
reidmix: Nice, very simple.  

If you wanted to be really expressive, you could have a helper called `has_message?` to return `@match[:message]`

lib/chat_ops/commands/timeline_command.rb
joshuatobin: +1 on log, i'd say leave `/t t` as well though.

lib/chat_ops/commands/end_command.rb
jmervine: I wanted to `/t x`, but I was trying to stay consistent with using the first letter.

lib/chat_ops/commands/timeline_command.rb
jmervine: Was going to `show` as well.

joshuatobin:
Also with regards to the slack command name, we can setup an alias on slacks side.

jmervine:
Huge +1 for moving from `/t` to `/incident` and `/ic`.

jmervine:
@joshuatobin @reidmix addressed comment and updated based on standup discussion.

app/controllers/api/v1/webhooks_controller.rb
```diff
@@ -29,7 +29,7 @@ def allowed_update_type?
   end
 
   def start_incident
-    ChatOps::Commands::StartIncidentCommand.process(nil, "start incident #{check_params[:incident_id]}")
+    ChatOps::Commands::StartCommand.process(nil, "start #{check_params[:incident_id]}")
```
joshuatobin: @jmervine this doesn't seem to be working correct. Also we need to fix the error message. 

screen shot 2016-12-06 at 3 15 21 pm

joshuatobin:
🌮 🎉 👍

lexelby:
Also a huge +1 from me to move to `/incident` / `/i`.

lexelby:
What about legacy Heimdall support?  I'm pretty sure this code messes heimdall up.

The (current) heimdall legacy support script requests a regex of all commands to match from retrodot.  Retrodot constructs the regex by concatenating all of the regexes from all command classes.  By adding the short/single-char aliases, we're polluting heimdall's namespace.  By changing "start incident" to just "/i start", we're also making it so the heimdall command is "h start".  That breaks the legacy command "h start incident" and also stomps on anything else in heimdall that might be "start <something>".

Maybe we should scratch the idea of concatenating the regexes together and just have the heimdall legacy support script translate commands into the new style before sending them to retrodot?

cc @jmervine @joshuatobin @reidmix

jmervine:
@joshuatobin @reidmix and I discussed this with @peculiaire and unanimously decided that this was a good time to toss out legacy support, especially because we're changing the slash command anyway. We decided that Retrodot should be considered a new thing, or at very least a major version release, where legacy support shouldn't be considered.

reidmix:
As I understand it, Lex is suggesting support for both.  This is the case for supporting the old `heimdall` way which now is just shims into the chatops command API.  We decided to support this for several months to aid in the transition.   Its a good catch that these changes will break `heimdall`.  IIUC, the suggestion is to do the translation at the `heimdall` shim layer to make sure everything is still working for `heimdall` as well as the streamlining that `/incident` does.

joshuatobin:
> Maybe we should scratch the idea of concatenating the regexes together and just have the heimdall legacy support script translate commands into the new style before sending them to retrodot?

@lex yeah that is probably the best idea

joshuatobin:
@lexelby can you update the coffeescript for this?

lexelby:
I can, but not immediately.  I'll try to get to it sometime this week.

PR #129, opened 2016-12-05
disabling db debug logging

jmervine:
@reidmix @joshuatobin


config/environments/production.rb
```diff
@@ -47,6 +47,10 @@
   # when problems arise.
   config.log_level = :debug
 
+  config.after_initialize do
+    ActiveRecord::Base.logger.level = 1
```
reidmix: Woot.  I'd say use `Logger::INFO` (which is 1) because it's more expressive.

joshuatobin: Can we make this a config var instead and default it to `Logger::INFO`?

jmervine: +1, this will allow us to easily enable debug logging in staging and such.

joshuatobin:
👍

config/environments/development.rb
```diff
@@ -56,6 +56,10 @@
     password:       Config.smtp_password
   }
 
+  config.after_initialize do
+    ActiveRecord::Base.logger.level = ("Logger::#{Config.log_level.upcase}".constantize rescue 0)
```
reidmix: I'd say, even with the `rescue 1/0` use the expressive names `Logger::Debug`/`Logger::Info`

jmervine: Okay, fine. xP  I was just  trying to keep the lines shorter.

PR #130, opened 2016-12-05
removing inspect around datetime for start and end commands

jmervine:
@joshuatobin This is done per https://trello.com/c/y2GnDP3O/218-look-into-remove-the-inspect-from-start-chatops-command

There's no visible difference, so not sure what the goal is here.

screen shot 2016-12-05 at 11 36 56 am

If it's to remove the syntax highlighting, then we'll also need to drop the back ticks.


reidmix:
Haha, that's awesome solution.  

For me the `.inspect` is a debugging feature, I'm not a fan of it being used outside of that.

reidmix:
👍

joshuatobin:
@jmervine The goal as Reid noted was just to remove the `.inspect`. It might have been an artifact from development. This LGTM.

👍

lexelby:
@jmervine @reidmix

When I wrote and tested this, there definitely was a visible difference.  Just interpolating the datetime into a string gave a numeric representation, while using inspect gave a more human-readable representation.

Here's what we have now in the heimdall implementation: `Tue, 13 Dec 2016 17:09:26 UTC`

`parse_timestamp` returns a `ActiveSupport::TimeWithZone`.  It behaves differently than a `Time` with regard to the behavior of `inspect`:

```
[30] pry(main)> t = Time.zone.now
=> Tue, 13 Dec 2016 19:35:00 UTC +00:00
[31] pry(main)> puts "#{t}"
2016-12-13 19:35:00 UTC
=> nil
[32] pry(main)> puts "#{t.inspect}"
Tue, 13 Dec 2016 19:35:00 UTC +00:00
=> nil
```

```
[33] pry(main)> Time.now.class
=> Time
[34] pry(main)> t = Time.now
=> 2016-12-13 14:35:29 -0500
[35] pry(main)> puts "#{t}"
2016-12-13 14:35:29 -0500
=> nil
[36] pry(main)> puts "#{t.inspect}"
2016-12-13 14:35:29 -0500
=> nil
```

We definitely need the more human-readable form.  People shouldn't have to parse a numeric datetime in their head during an incident, and the day-of-week is useful too.  If we want to avoid `inspect`, we should use `strftime`.

reidmix:
oh interesting,  I remember this one and  I asked for the difference, it was exactly the same.  I wonder what happened?  Anyway, I believe `#rfc822` is what you are looking for.

lexelby:
Yeah, it's very subtle -- you only see a difference if the timestamp is of type `Time.zone.now`.

I remember testing rfc822 too:

```
[5] pry(main)> t = Time.use_zone(Config.time_zone) do Time.zone.now end
=> Tue, 13 Dec 2016 13:16:47 PST -08:00
[6] pry(main)> t.rfc822
=> "Tue, 13 Dec 2016 13:16:47 -0800"
[7] pry(main)> t.inspect
=> "Tue, 13 Dec 2016 13:16:47 PST -08:00"
```

I chose `.inspect` because it gave the 3-letter representation of the time zone.  But I also didn't know that `inspect` isn't generally used in production code. :)  So I guess either we accept `-0800` or we use `strftime`.

PR #131, opened 2016-12-06
Collapse attachments

joshuatobin:
Prepending to the attachment will force a collapse even if the attachment is 1 line. 

Before - it would collapse no matter what.

screen shot 2016-12-06 at 10 02 36 am

After:

screen shot 2016-12-06 at 9 55 09 am

CC: @jmervine @reidmix


jmervine:
👍

reidmix:
I suggest to put the "Show more..." on its own line so it's easy to see/click.

lexelby:
@reidmix @joshuatobin 

It was _purposeful_ that the timeline collapses without showing any lines.  That's why I prepend in the heimdall incident response code.  My goal is to avoid cluttering #platform-incidents as much as possible.  If individuals (such as the person that requested the timeline) want to expand the text, they can.

My vote is to revert this PR.

lexelby:
Also, FWIW, we don't get the option of putting the "show more..." on its own line.

jmervine:
Most of these PRs came out of feedback from user testing. I believe this is one of those where people preferred it collapsed. I'm not certain about that on this one though.

joshuatobin:
@jmervine is correct. I'm -1 on reverting.

PR #132, opened 2016-12-06
Return a list of responders when adding/removing from timeline

joshuatobin:
CC: @jmervine @reidmix


jmervine:
👍

lexelby:
Was this something that came out of user testing?  I didn't return the full list of responders originally to avoid cluttering the channel, but I could see how that would be confusing or not useful to users.

jmervine:
Not sure if this came out of user testing or not. Personally, I was more in favor of this when responders were all on one line, but not that they're expanded, it might be a little much. Perhaps return a single line list of responders in this case and leave the expanded list for `/i responders`?

joshuatobin:
This came out of user testing as well.

lexelby:
Gotcha, so people wanted to see the list returned after they add or remove.  Perhaps it makes sense to send the list as it is now, using the "only you can see this" method, or just on one line like @jmervine suggests.

PR #133, opened 2016-12-06
Fix inline styling for retro emails 

joshuatobin:
Fixes broken body tag, coloring entire email red. `body -> strong tag`. Tested in apple mail as well.
screen shot 2016-12-06 at 12 44 51 pm

CC: @reidmix @jmervine


jmervine:
👍

PR #135, opened 2016-12-07
Fix user rake tasks

joshuatobin:
This is a fix for a race condition in the users rake tasks. `Mediators::User:Sync` is setting a handle attribute and then `Mediators::User::SyncSlackUsers` comes along and nil's out the values for the handle. This fix will allow `User.ensure` to run and not update users with nil values.

Also keep in mind `Mediators::User:Sync` is a bandaid and we'll be removing this post MVP. I've also made a change to `Mediators::User:SyncSlackUsers`, in which `handle` is actually just referred to as `name`, and `name` is referred to as `real_name` on slacks side.

Lastly this also fixes another bug which we've ran into a few times: "Slack `user_id` not found". Again from the rake task nilling out the id, and then the next rake task filling it back in. 

CC: @jmervine @reidmix


jmervine:
👍

lexelby:
@joshuatobin Actually, now that we have `Mediators::User::SyncSlackUsers`, we don't need `Mediators::User::Sync` at all.  `Sync` was because we didn't initially plan for retrodot MVP to access slack directly at all.

PR #136, opened 2016-12-07
admin command

jmervine:
@joshuatobin @reidmix as discussed

updated generators based on changes made since originally implemented


joshuatobin:
👍

PR #137, opened 2016-12-07
Return a warning for previosuly started incidents

joshuatobin:
CC: @jmervine @reidmix


jmervine:
👍

lexelby:
@joshuatobin Why?? People adjust the start time for incidents all the time.  They also occasionally start the wrong incident, especially because `/i start` tries to auto-detect which incident you want to mean if you don't specify the ID.

jmervine:
The idea was to simplify for MVP. The current work flow was confusing (even in Hemidall), where start times would get accidentally updated ... I know it happened several times in user testing and I used to do it all the time when IC.

I want to add the concept of "`!`" commands, i.e. `start!` and `end!` for modifying start and end, post MVP, for now though, it was decided this wasn't necessary, as you should be able to update this value though the admin interface.

lexelby:
Well, this does mean we're _reducing_ functionality for the MVP.  Perhaps add a hint that they can edit the start time in the retrodot web UI?

jmervine:
If the current functionality is broken -- not functionally, but from a UX standpoint. Then I'd say yeah.

joshuatobin:
The current implementation isn't clear. We confirmed that in user testing, not only from others but people directly on the team so I feel reeling back in for now is a good plan. This IMO is reducing confusion, not functionality. We can focus specifically on how to do it differently in our upcoming sprint.

lexelby:
I see the point.  It is pretty verbose and confusing as it is now.  But I do know that some users have updated start times in the past.  Perhaps we can at least socialize that the retrodot admin UI allows changing timeline start time?

PR #141, opened 2016-12-08
user testing feedback loop 1

jmervine:
- [added `/ic incidents` command](https://trello.com/c/GUBbOXtQ/252-command-for-previous-10-incidents)
- [removed single letter shortcuts for matching bug](https://trello.com/c/edoDdS3S/254-bug-ic-list-became-a-shortcut-for-adding-a-timeline)

@joshuatobin @reidmix


joshuatobin:
👍

lib/chat_ops/commands/incidents_command.rb
```diff
@@ -0,0 +1,17 @@
+module ChatOps::Commands
+  class IncidentsCommand < ChatOps::Command
+    setup do
+      match(/^(incidents|ids)($|\s+)((?<limit>\d+)?)$/)
```
reidmix: how about `/incident recent` instead of `/incident incidents?

lib/chat_ops/commands/incidents_command.rb
```diff
@@ -0,0 +1,17 @@
...
+      help "`incidents` list the past 10 incidents"
+    end
+
+    def run
+      limit = 10
```
reidmix: Config var?

lib/chat_ops/commands/incidents_command.rb
reidmix: Do you think `list` on this command might be a better fit?

joshuatobin: Good call on `recent`? I worry we're gonna overload the usage of`list`.

lib/chat_ops/commands/incidents_command.rb
joshuatobin: 10 seems like a reasonable default but /shrug.

lib/chat_ops/commands/incidents_command.rb
```diff
@@ -0,0 +1,17 @@
...
+      incidents = ::Incident.limit(limit).map do |i|
+        "`#{i.incident_id}` #{i.title}"
```
jmervine: I was thinking link in to status.heroku.com, but this has merit. I think it's something we'd have to ask for user feedback on once it's been live for a while.

PR #143, opened 2016-12-08
adding ping command

jmervine:
addressing https://trello.com/c/fafnhFGh/257-add-ic-ping-to-ping-all-responders

@reidmix @joshuatobin


jmervine:
WIP, @ mentions aren't working correctly

lib/chat_ops/commands/ping_command.rb
```diff
@@ -0,0 +1,18 @@
+module ChatOps::Commands
+  class PingCommand < ChatOps::Command
+    setup do
+      match(/^(ping|at)((\s+(?<incident_id>\d+))?)\s+(?<message>.+)*$/)
```
reidmix: :-1: on "at" because I'm not sure I understand that name.

`/ic ping everyone give me their status`
vs
`/ic at everyone give me their status`

jmervine: yeah, wasn't super sure about it either

lib/chat_ops/commands/ping_command.rb
```diff
@@ -0,0 +1,18 @@
...
+      help "`ping [#] <message>` send a message to all responders"
+    end
+
+    def run
+      responders = @incident.responders.sort.map do |responder|
+        handle = responder.handle
+        handle = "@#{handle}" unless handle.starts_with?("@")
```
reidmix: doesn't all the `responders`'s handle come without an `@` sign?  I'm guessing yes.
In that case you could have a helper:
```ruby
def at_handle
  "@#{handle}"
end
```

lib/chat_ops/message_helpers.rb
```diff
@@ -14,6 +14,10 @@ def message(text, **args)
       { message: prevent_highlights(text), **args }.compact
     end
 
+    def highlight_message(text, **args)
+      { message: text, **args }.compact
```
joshuatobin: <3 thanks 💃

lib/chat_ops/commands/ping_command.rb
joshuatobin: Yeah they should be normalized in the db from the `SyncSlackUsers` Mediator, so you shouldn't need the `starts_with?`

joshuatobin:
@jmervine Nice work on this 👍

joshuatobin:
CC: @slant too as it was his idea 💃

reidmix:
👍

lexelby:
Ooh, can't wait to use this.  Will let us avoid the need for @here.

lexelby:
Errm, sorry about that @-mention, Mr. Dubs.

PR #144, opened 2016-12-09
Slack event bug fixes

joshuatobin:
There are a slew of bugs in the ChatOps API controller, mostly around`slack_events`. This is one of two PRs. The second PR will address messages getting added to the timeline will be from the originating author of the message, *not* the user that reacted to it. 

Bugs:
* None of the locales were working and failing silently without the include. The `t()` helper is only available in rails views automagically
* All of the renders were returning non 200 statuses, which is normally fine, but this meant slack just kept sending us retries. No bueno. 
* Added logic so that if someone reacts to a message originating from the bot, that message is discarded. 
* Added more tests 

CC: @jmervine @reidmix


jmervine:
Nice! Might be worth mentioning that there's also a slew of text throughout the commands code that could use being moved to the locales file. Perhaps not in this PR though.

👍

config/locales/en.yml
```diff
@@ -28,6 +28,7 @@ en:
       incident:
         missing: "No incident currently in progress"
       reaction: "Invalid timeline reaction"
+      bot_reaction: "Error adding message to timeline. Messages originating from the bot are not supported."
```
reidmix: You could be whimsical :D "Hey, don't listen to me" etc. :D

app/controllers/api/v1/chat_ops_controller.rb
```diff
@@ -35,28 +37,38 @@ def slack_slash_command
...
-    end
+    return slack_error_to_user(error: t("slack.errors.incident.missing")) unless Incident.in_progress?
 
-    entry = add_timeline_entry(slack_message_from_timestamp, Incident.open_incident_id)
-    post_to_slack(entry[:message], slack_channel)
+    entry = add_timeline_entry(slack_message_from_timestamp["text"], Incident.open_incident_id)
+    post_to_slack(entry[:message], slack_channel_from_event)
 
-    render json: t("slack.timeline.added"), status: 201
+    render json: t("slack.timeline.added"), status: 200 # Render 201 so slack doesn't retry
```
reidmix: Not sure I understand the comment with the `200`

PR #145, opened 2016-12-09
Add timeline entries originating from the author, not the user that reacted

joshuatobin:
This will add the originating user to the timeline vs the person who reacted. NOTE: I was a bit verbose in the response messaging to make it entirely clear that if someone did click on a reaction and it adds a message to the timeline by mistake, they should have a pretty good idea of why they're getting a notification.

CC: @jmervine @reidmix


joshuatobin:
Private message send to slack bot: 
screen shot 2016-12-09 at 2 45 48 pm

joshuatobin:
CC: @peculiaire As i think this was your request.

jmervine:
@joshuatobin This is better than it is, but it still leaves a question as to what was reacted. Perhaps instead of saying "Message successfully added...", include the message in quotes, like '"<message>" has been successfully added...'

Perhaps overkill, but just a thought while you're in there.

joshuatobin:
@jmervine Good suggestion. I've updated the response message to be:

screen shot 2016-12-12 at 10 10 16 am

jmervine:
👍

PR #147, opened 2016-12-13
Add timeline entries with the original timestamp via slack reactions

joshuatobin:
Rather than trying to access `ChatOps::Command` from the controller and having to hack in the regex logic we're going directly to the model to create the entry with the timestamp. 

CC: @jmervine @reidmix


app/controllers/api/v1/chat_ops_controller.rb
```diff
@@ -49,9 +49,12 @@ def slack_event
 
     # Add a timeline entry from the originating author of the message
     # and respond to the user who reacted to the message with a private confirmation
-    entry = add_timeline_entry(slack_user_from_message_ts, slack_message_from_timestamp["text"], Incident.open_incident_id)
-    entry = [entry[:message], timeline_success_msg]
+    current_incident.timeline_entries.create!(
```
jmervine: This will throw `NoMethodError: undefined method `timeline_entries' for nil:NilClass` if there is no `current_incident`. Might want to rescue and report back to the user that there's no current incident, pretty sure there's a helper for that.

spec/controllers/api/v1/chat_ops_controller_spec.rb
```diff
@@ -115,14 +117,21 @@
...
         expect(response.body).to match(/Timeline entry successfully added/)
       end
 
+      it "adds the proper timestamp for timeline entries via reaction" do
+        allow_any_instance_of(Slack::Client).to receive(:channels_history).and_return(slack_message_from_history)
+        expect(User).to receive(:find_by).and_return(user)
+
+        post :slack_event, params: event
+        expect(incident.timeline_entries.first.timestamp.utc).to eq Time.at(slack_timestamp).utc
+      end
```
jmervine: Might also want to consider adding a test case for when there's no incident.

app/controllers/api/v1/chat_ops_controller.rb
joshuatobin: I'll add a test but it should never get this far since line 48 handles the case of no `current_incident`

spec/controllers/api/v1/chat_ops_controller_spec.rb
joshuatobin: Will do.

joshuatobin: @jmervine actually this should cover it: https://github.com/heroku/retrodot/blob/master/spec/controllers/api/v1/chat_ops_controller_spec.rb#L83

jmervine:
👍

PR #149, opened 2016-12-14
Add .profile to properly set TERM in heroku spaces

joshuatobin:
CC: @jmervine @lexelby @reidmix tested via `run:inside`


reidmix:
👍

PR #150, opened 2016-12-14
compact responders on add/remove

jmervine:
Based on conversations with @lexelby around responders formatting. Compacting responders to a single line on `add` and `remove`.

cc @reidmix @joshuatobin @lexelby


lib/chat_ops/commands/add_command.rb
```diff
@@ -11,7 +11,7 @@ def run
       return error("Person not found.  Try using their @handle.") if responders.empty?
 
       @incident.responders += responders
-      return_responders
+      return_responders compact: true
```
joshuatobin: Its a nitpick but i'd prefer `()` here

joshuatobin:
👍

lib/chat_ops/commands/add_command.rb
jmervine: done

PR #157, opened 2016-12-20
Adding oauth scope for events api

joshuatobin:
According to the [events api documentation](https://api.slack.com/events-api), we need the `reactions:read` [scope](https://api.slack.com/docs/oauth-scopes). Its possible they've recently updated our scopes since this works fine in staging. 

CC: @jmervine @lexelby


jmervine:
👍

Issue #154, opened 2016-12-20
Consider having '/ic log' unfurl images in it's reply...


jmervine:
<img width="581" alt="screen shot 2016-12-20 at 11 50 12 am" src="https://cloud.githubusercontent.com/assets/837893/21365537/8bf4e73e-c6aa-11e6-9055-edc35dc3dfac.png">

jmervine:
This is actually an issue with slack and how it prints when using a `/` command.

Issue #155, opened 2016-12-20
Incident start and end times to the timeline.


jmervine:
It'd be rad if the start and end were logged in the incident timeline, both when doing `/ic timeline` and in the google doc.

<img width="637" alt="screen shot 2016-12-20 at 12 10 07 pm" src="https://cloud.githubusercontent.com/assets/837893/21366114/42f1b884-c6ad-11e6-97dd-1fc9a88d49d0.png">

jmervine:
Addressed in #159

PR #159, opened 2016-12-20
adding start/end message to timeline

jmervine:
addressing #154 

When incident is started without a user (Events API or push from status I'm guessing?) I've allowed this to add the timeline message anyway. I'm think catching unknown users in `timeline_command.rb` to ensure that an error isn't thrown, but rather "Unknown" is reported as the user.

I'm thinking a better way of handling this would be to create an "Events API" user, or "Status Site" user and then use that as the user when starting an incident via automation. Something like `User.create(name: "Events API", handle: nil, email: "sre-team-only@heroku.com")`. Discuss.


lib/chat_ops/commands/end_command.rb
```diff
@@ -8,9 +8,12 @@ class EndCommand < ChatOps::Command
 
     def run
       @incident.chat_end = Time.zone.now
+
+      message = "Recorded the end of chat for incident `#{@incident.incident_id}` at *#{@incident.chat_end}*."
+      @incident.timeline_entries << ::TimelineEntry.new(user: @user, message: message)
```
joshuatobin: 👍

lexelby:
Looks great!  I wonder if it makes sense to allow timeline entries to have a `nil` user field instead?

lexelby:
:+1:

joshuatobin:
I like the idea of having an "Events" and "Status" user.

jmervine:
Currently, they'll have a `nil` user field. Also, it's not looking like there's actually any occurrence were the "start" command is run without a user. As such, I'm going to merge this down. I do want to discuss the idea of creating canned users for events api and status site.

reidmix:
Issue #155 btw <3

reidmix:
I'm also a +1 on the Status user to show where it came from.  That's a fine idea.

PR #161, opened 2016-12-21
disable rollbar in rails console

lexelby:
People typo all the time in the rails console and it ends up sending an email with a rollbar error.  It can be confusing and lead others to do pointless debugging.  Let's not send any top-level exceptions to rollbar when in the rails console.  A developer is already there and will see it anyway.


jmervine:
👍 We should probably roll up quick PRs for Laika, Status, W:TNG and the rest with this in there too.

jmervine:
merging, as I'm going to do a release today

lexelby:
Sure, although I should note that I haven't tested this at all.  I don't think it'll break anything but I'm not certain it'll do what I intended :)

Issue #160, opened 2016-12-21
remove "sync:users" task


lexelby:
This was deprecated in favor of syncing from slack before we even deployed retrodot to production.

jmervine:
Removed with #162

PR #162, opened 2016-12-21
removing sync:users

jmervine:
addressing #160 

@lexelby did I miss anything?

When releasing, remove `CHATOPS_USERS_URL` and `CHATOPS_USERS_API_KEY` config settings from environments.


lexelby:
Config var.  Other than that, I don't think so. :)

jmervine:
@lexelby Removed the config vars.

lexelby:
:+1:

PR #164, opened 2016-12-21
readding factory spec fix

jmervine:
The factory change must have accidentally been removed in a merge somewhere. It makes the memorization of the process user unnecessary.

cc @lexelby


lexelby:
:+1:

PR #171, opened 2017-01-03
fix small issue with prepare retro

lexelby:
I wish we didn't have to test in staging.

I've tested this in staging (incident 1006) and it works.  Doc is prepared and responders and timeline are filled in properly.


lexelby:
@joshuatobin @jmervine 

Feel free to merge this evening or I'll do it tomorrow when I get in.

jmervine:
:+1:

reidmix:
Can you think of a way we don't have to test in staging?  Could we figure out how mock out Google/Trello/Slack?  Are there other reasons?

lexelby:
One issue is that we have to give Google a whitelist of URLs that they can redirect to after an oauth request is approved by the user.  From what I can tell, google doesn't allow wildcards for those URLs, so we have to add one manually in the google developer console for each new PR app.

Another problem is that the `/i` command in our testing slack is hooked up to retrodot-staging.  We'd have to add a new /-command for each PR app every time we wanted to test.  It's a colossal pain.

The best solution currently seems to be to force-push directly to the retrodot-staging heroku app BEFORE merging the PR and test in staging.  Then revert retrodot-staging, get the PR app approved, and merge it, which will auto-deploy to staging via the normal pipelines system.

lexelby:
Another possibility, I suppose, would be to add a `retrodot-development` app, set it up in google and slack, and then do our force-pushes to it to test PRs, thus keeping staging clean.

joshuatobin:
@lexelby Yeah +1 on the `retrodot-development` app. Good idea.

reidmix:
`retrodot-development` may be the way to go. 

I'm curious how the tools team deals with these items?  I think they'd have the same scenario.

Also on mocking (or in general), I wonder if people in the community have built mocks for google/trello/slack or if they have best practices too?

lexelby:
A mock isn't really helpful when the thing we're trying to test is whether the actual google or slack stuff works, though...

PR #172, opened 2017-01-04
Delay response to /-commands

lexelby:
This works around a slack client bug.  See the code comments for details.


app/controllers/api/v1/chat_ops_controller.rb
```diff
@@ -22,18 +22,27 @@ def respond
...
     return render_slack_error(t("slack.errors.user.id"), 403) unless user
 
     @result = ChatOps.process(user, chat_params[:text])
 
+    # The slack client has a bug that is triggered by a /-command response
+    # coming too quickly.  Chat doesn't scroll down to show the response.
+    # To work around this, make sure we take at least half a second to respond.
+    sleep_time = 0.5 - (Time.zone.now - start_time)
+    sleep sleep_time if sleep_time.positive?
+
```
joshuatobin: @jmervine Didn't you try something like this pre MVP and it didn't workout?

lexelby:
Doesn't work.  Retrodot responds in such a way as to request that slack echo the /command in the channel.  The /command is echoed right before retrodot's response, no matter how long we delay our HTTP response, and this triggers the scrolling bug.  We would need to return a blank response to have the command echoed and then post the actual result to the channel separately in an async job.

joshuatobin:
@lexelby Might it be worth seeing if we can escalate this up to slack again?

lexelby:
Sure, I'll give them a poke.

lexelby:
I heard back from them:

> Hi Lex,
>
> Sorry for the wait time there! It looks like our engineers are still working on this bug, as we've been able to reproduce this as well. I really wish I had better news at this point! I can see how this failure to scroll is really annoying, and I really hope we can get that resolved sooner than later. I've linked up this ticket to our bug tracking system, so that as soon as the engineers have an update, I'll be able to let you know promptly.
> 
> My apologies again for the frustration here, Lex. If I can help with anything more in the meantime, please let me know.
> 
> Best,
> 
> Megan

This is useful because it means I'll get an email when they've fixed it and we can remove our workarounds.

cc @reidmix

joshuatobin:
👍 on the response

PR #173, opened 2017-01-05
queue async job to reply to /-commands after delay

lexelby:
redo of #172.

Instad of replying to slack's HTTP request with the full response, now we save it and POST it to the URL they provide.  The hardest part of this was figuring out how to test ActiveJob.

Currently running on staging.  Tested and works :D

cc @joshuatobin @jmervine


joshuatobin:
👍

Issue #167, opened 2016-12-21
ChapOps API Error in error handler...


jmervine:
[This block](https://github.com/heroku/retrodot/blob/master/app/controllers/api/v1/chat_ops_controller.rb#L64-L69) is causing an uncaught, unsuspected, error b/c `error` is not a valid Exception object.

Consider something like

```ruby
rescue Slack::Error
    error = Slack::Error.new("#{t('slack.errors.timeline')}: #{$ERROR_INFO}")
    post_to_slack(error.message, slack_user_from_event)
    log_error(error: error)
    render json: error, status: 200 # Render 201 so slack doesn't retry
end
```

Or perhaps just 

```ruby
rescue Slack::Error => error
    error.message = "#{t('slack.errors.timeline')}: #{$ERROR_INFO}"
    post_to_slack(error.message, slack_user_from_event)
    log_error(error: error)
    render json: error, status: 200 # Render 201 so slack doesn't retry
end
```

lexelby:
Or we could just do `log_error(error_message: error` to avoid loggerator's exception printing logic.

jmervine:
Wasn't familiar with that. 👍🏽

joshuatobin:
Closing as this has been merged.

PR #178, opened 2017-01-06
Fixing error logger

jmervine:
addresses #167 

Docker related changes are to support `command` handling in newer version of `docker-compose` as well as make `docker-compose up` work correctly.


joshuatobin:
👍

Issue #176, opened 2017-01-06
Feature Request: Timeline reaction should add to last incident if no incident is currently running


peculiaire:
I would like to be able to add stuff to the timeline of a previous incident via reaction. 

User story:

Owen comes into P-I and notifies us that we had a build outage. It's since been resolved. IC/Comms post a retroactive incident. IC adds timeline reaction to chat to build a timeline for the retroactive incident.

lexelby:
I wondered if there might be a use case for that.  Thanks for the detailed description!  Should be low-effort.

joshuatobin:
@lexelby @jmervine Should just be a tweak to https://github.com/heroku/retrodot/blob/master/app/controllers/api/v1/chat_ops_controller.rb#L55

Thoughts on ill effects?

lexelby:
At worst (and unlikely) this attaches a timeline entry to the wrong incident.  People can edit timelines in their google docs, so I feel like it's low-impact.

jmervine:
I think there's a trello card for this. I know I called it out in user-testing. Not just adding to timeline, but for the following commands as well...

```
/i who
/i timeline
/i add <person>
/i rm <person>
```

jmervine:
We should also update the admin interface as part of this to allow for editing responders and timelines.

joshuatobin:
Closing as this has been merged

PR #188, opened 2017-01-10
Add timeline reactions to the last incident

joshuatobin:
As requested in https://github.com/heroku/retrodot/issues/176

CC: @reidmix @peculiaire @lexelby @jmervine


app/controllers/api/v1/chat_ops_controller.rb
```diff
@@ -78,7 +74,7 @@ def slack_event
   private
 
   def current_incident
-    Incident.by_timeline_start.first
+    Incident.first
```
lexelby: I think this should still be `Incident.by_timeline_start.first`.  We want it to pick the incident most recently started, just like with chatops commands like `/i timeline foo`.

spec/controllers/api/v1/chat_ops_controller_spec.rb
```diff
@@ -72,6 +72,19 @@
       end
     end
 
+    describe "Closed Incident" do
+      let!(:incident) { create(:incident, timeline_start: Time.now.utc, open: false) }
+      it "uses the last incident" do
```
lexelby: This may not be worth the effort, but is it possible to verify that the message was actually added to the right incident's timeline?

lexelby:
:+1:

I'm okay with leaving the test as it is if it'd be too much effort to make it more rigorous. Your call.

spec/controllers/api/v1/chat_ops_controller_spec.rb
```diff
@@ -72,6 +72,21 @@
...
+      let!(:now) { Time.now.utc }
+      let!(:incident) { create(:incident, timeline_start: now, open: false) }
+      it "uses the last incident" do
+        allow_any_instance_of(Slack::Client).to receive(:channels_history).and_return(slack_message_from_history)
+        expect(User).to receive(:find_by).and_return(user)
+        expect(Incident.in_progress?).to be false
+
+        post :slack_event, params: event
+        expect(response).to have_http_status(:ok)
+        expect(incident.timeline_start).to eq now
```
joshuatobin: @lexelby tested it by verifying the `timeline_start` is the same.

lexelby: Ah, nice!  I'd recommend you wrap the whole test in `Timecop.freeze do ... end` to avoid having your test randomly fail if it happens to run _right_ as a second ticks over to the next.

spec/controllers/api/v1/chat_ops_controller_spec.rb
```diff
@@ -73,17 +73,19 @@
...
-      it "uses the last incident" do
-        allow_any_instance_of(Slack::Client).to receive(:channels_history).and_return(slack_message_from_history)
-        expect(User).to receive(:find_by).and_return(user)
-        expect(Incident.in_progress?).to be false
-
-        post :slack_event, params: event
-        expect(response).to have_http_status(:ok)
-        expect(incident.timeline_start).to eq now
-        expect(response.body).to match(/Timeline entry successfully added/)
+      Timecop.freeze do
```
lexelby: Sorry, I should've been clear.  This needs to be inside the `it` block.

PR #192, opened 2017-01-11
Properly catch the exception, without the named param

joshuatobin:
Since we were using a named param: `error:`, the loggerator gem was blowing up underneath, as it was just excepting a regular param to be passed. 


lexelby:
:+1:

reidmix:
:+1:

PR #193, opened 2017-01-11
Don't enqueue ephemeral messages 

joshuatobin:
This fixes a bug from slackbot:

screen shot 2017-01-11 at 2 10 57 pm


lexelby:
:+1:

PR #196, opened 2017-01-16
Log update type

joshuatobin:


reidmix:
My thought: make sure there's no auth/tokens/etc. -- i don't think there is, but there could be in the future maybe?  There is this: https://github.com/heroku/rollbar-blanket

joshuatobin:
@reidmix yeah good point. I'm just gonna go back to my original plan. Please look again

lexelby:
:+1:

jmervine:
👍

PR #198, opened 2017-01-16
Follow redirects

joshuatobin:
Found that retrodot wasn't following redirect, and in turn breaking webhook functionality. 


jmervine:
👍

lexelby:
Ideally we'd just have retrodot hit the new status URL directly rather than go through the redirect, but I'm okay with adding redirect support, sicne it can't hurt.

PR #200, opened 2017-01-17
Refactor webhooks

joshuatobin:
I've refactored the controller to include the following:

* Post to slack for incidents starting a timeline
* Post to slack for resolved incidents
* Condensing the sync method.
* Better error handling / logging.


app/controllers/api/v1/webhooks_controller.rb
```diff
@@ -2,35 +2,59 @@
...
+  def sync_incident
+    Mediators::Incident::OneSyncher.run(id: incident_id)
+  end
+
+  def process_webhook
+    @message =
+      if scheduled_maint?
+        "update_type: '#{update_type}' received."
+      elsif resolved?
+        end_incident
```
lexelby: We didn't auto-close incidents in heimdall, and I hadn't yet done it that way in retrodot.  My reasoning was that chat for an incident often continues on past when we resolve it on the status site.  My original thinking with the start/end times is that retrodot could theoretically fetch the entire chat log for an incident from slack -- though we don't do that currently.

Anyway, that's why I vote we don't auto-close stuff in retrodot.  Closing means people will then have to specify the incident ID if they want to add more stuff to the timeline, too.  On the downside, people might forget to `/i end`.  So I don't feel very strongly about my position on this.

joshuatobin: @jmervine and I talked about having a few commands work when incidents are resolved, by automagically picking the last incident. I think that would satisfy your concern. Either way I think people aren't going to remember to end the incident and it should be automated.

app/controllers/api/v1/webhooks_controller.rb
```diff
@@ -2,35 +2,59 @@
...
-  def can_start_incident?
-    allowed_update_type? && @incident && @incident.missing_timeline?
+  def sync_incident
+    Mediators::Incident::OneSyncher.run(id: incident_id)
+  end
+
+  def process_webhook
+    @message =
+      if scheduled_maint?
+        "update_type: '#{update_type}' received."
```
joshuatobin: Wondering if we should have scheduled maint start an incident and post to slack?

app/controllers/api/v1/webhooks_controller.rb
joshuatobin: I just realized, this might blow up if we're trying to resolve a scheduled maint. Need to test this more.

lexelby: I agree, people generally won't remember to resolve it.  I'm a big +1 on having commands automatically pick the last incident.

Also, looking back on this, if my concern is that some kind of chatlog fetch would miss messages that happened after an incident, we could just have the code be "smart" and fetch messages that continued after until people stopped chatting :)

app/controllers/api/v1/webhooks_controller.rb
lexelby: Here's how I think it should work.  _Scheduling_ the maintenance shouldn't start an incident.  A scheduled maintenance also should not be considered when one does a plain `/i start` -- retrodot should never assume you meant a scheduled maintenance.  Once the actual maintenance starts, the support person clicks "begin maintenance" in status admin and status adds a post of type "maintenance" (I think).  _That_ one should start the timeline in retrodot.  That way if things to pear-shaped people can add timeline messages.

app/controllers/api/v1/webhooks_controller.rb
```diff
@@ -2,35 +2,59 @@
...
+    log_error(e, at: :sync, incident_id: incident_id, update_type: update_type)
+    render json: t("slack.errors.webhook.incident"), status: 500
   end
 
   private
 
-  def can_start_incident?
-    allowed_update_type? && @incident && @incident.missing_timeline?
+  def sync_incident
+    Mediators::Incident::OneSyncher.run(id: incident_id)
```
jmervine: Is `incident_id` defined elsewhere as a method? If not, `fetch_incident` could probably set a class var, and an accessor could be added.

app/controllers/api/v1/webhooks_controller.rb
```diff
@@ -2,35 +2,59 @@
...
-  # rubocop:disable AbcSize
-  def sync
-    @incident = fetch_incident
+  def sync # rubocop:disable AbcSize
     unless @incident
       log(at: :sync, incident_id: incident_id, message: "id not found. Syncing")
-      Mediators::Incident::OneSyncher.run(id: incident_id)
-      @incident = fetch_incident
+      sync_incident
+      fetch_incident
```
jmervine: Here in `sync`, you're running `sync_incident`, then `fetch_incident`, which appears to be running `sync` again with the `before_action` above. There's unnecessary looping here, in fact, it should probably be causing ruby to core dump (?).

Also, I would consider running `fetch_incident` early, and then using `@incident` to define the `id` and `update_type` throughout.

joshuatobin:
@lexelby Addressed your comments in regards to starting with update_type of `maintenance`

app/controllers/api/v1/webhooks_controller.rb
joshuatobin: I don't believe so. I'm running `fetch_incident` as a `before_action`, to see if there is an existing incident, is not, it runs the sycher, and tries to re-lookup the incident_id. So the lookup is done early, and needs to be run a second time to properly set @incident.

app/controllers/api/v1/webhooks_controller.rb
```diff
@@ -2,35 +2,63 @@
...
+  end
+
+  def maintenance?
+    check_params[:update_type] == "maintenance"
   end
 
-  def allowed_update_type?
-    !%w{scheduled maintenance}.include?(check_params[:update_type])
+  def end_incident
+    ChatOps::Commands::EndCommand.process(nil, "end #{incident_id}")
```
jmervine: I'd prefer avoiding running `process` commands manually, as discussed before. But rather surfacing the applicable code with in the command class and executing that, bypassing the regex, where possible. We should have the incident id already, so there's no need to parse it out again. In this case, I'd probably refactor `run` in the end command class to be...

```
    def run(incident:, user:)
      incident ||= @incident
      user ||= @user

      incident.chat_end = Time.zone.now

      message = "Recorded the end of chat for incident `#{incident.incident_id}` at *#{incident.chat_end}*."
      incident.timeline_entries << ::TimelineEntry.new(user: user, message: message)
      incident.save

      { message: message }
    end
  end
```

That's not going to work out of the box though, b/c we don't have a user in the webhook. So... now that I'm thinking about it, maybe leave it as is, but consider refactoring the commands for this kind of support and creating a webhook/status/unkown/whatever user as discussed before for these kinds of actions.

app/controllers/api/v1/webhooks_controller.rb
reidmix: Agreed. 

You're correct, starting a "scheduled" adds a "maintenance" update. 
Example: https://status.heroku.com/incidents/1021

app/controllers/api/v1/webhooks_controller.rb
joshuatobin: I agree with you but I'd rather not attempt to do this in this PR.

app/controllers/api/v1/webhooks_controller.rb
joshuatobin: @reidmix @lexelby I've fixed this. Can you pls look again?

app/controllers/api/v1/webhooks_controller.rb
```diff
@@ -2,35 +2,63 @@
...
+    start_incident? || resolved? || maintenance?
+  end
+
+  def sync_incident
+    Mediators::Incident::OneSyncher.run(id: incident_id)
+  end
+
+  def process_webhook
+    @message =
+      if maintenance?
```
lexelby: Can this condition be `maintenance? || start_incident?` so that you can avoid duplicating the `start_incident[:message]` line?

joshuatobin: yup sure

app/controllers/api/v1/webhooks_controller.rb
```diff
@@ -2,35 +2,61 @@
...
+  end
+
+  def sync_incident
+    Mediators::Incident::OneSyncher.run(id: incident_id)
+  end
+
+  def process_webhook
+    @message =
+      if maintenance? || start_incident?
+        start_incident[:message]
```
joshuatobin: @lexelby fixed!

lexelby:
:+1:

I agree with what both you and @jmervine said about refactoring the call into the chatops command: good idea but now's not the best time to do it.

Issue #179, opened 2017-01-06
Feature Request: autocomplete email addresses when sending retro email


reidmix:
Until we know the team manager & teams automatically, it'd be nice to have an auto-complete of email addresses when I'm sending the retro email similar to how mail does it.

lexelby:
+1

lexelby:
Oops.  Wrong button.

lexelby:
Doesn't look like there's anything close to an API for gathering this information or auto-completing email addresses.  Our best bet may be to open a gmail composition window, which handily solves the problem I had too, which is that sometimes I want to edit the message.  Check this out: http://stackoverflow.com/a/8852679/4249120

joshuatobin:
Maybe a long shot but we probably could have retrodot accept CC/BCC messages and save them to the db so we have a record of the thread?

lexelby:
I think that might be doable!  That'd be neat.  Probably should be a separate ticket though.

lexelby:
Ah, I see the problem: without having retrodot receive a CC/BCC, it won't know for sure that the retro email has been sent.

Another problem is that the URL technique described above for pre-filling a gmail compose window doesn't allow HTML.  That means we can't do things like the bold/red "This incident will need a public postmortem" message.  We could sort of do bold with unicode math bold characters like this:

𝗧𝗵𝗶𝘀 𝗶𝗻𝗰𝗶𝗱𝗲𝗻𝘁 𝘄𝗶𝗹𝗹 𝗻𝗲𝗲𝗱 𝗮 𝗽𝘂𝗯𝗹𝗶𝗰 𝗽𝗼𝘀𝘁𝗺𝗼𝗿𝘁𝗲𝗺 𝗯𝘆 𝟮𝟬𝟭𝟳-𝟬𝟭-𝟬𝟯.

That was generated using this tool: http://qaz.wtf/u/convert.cgi?text=This+incident+will+need+a+public+postmortem+by+2017-01-03.

Kind of hacky and I can't find a gem to do that kind of translation, so we'd have to write one ourselves or something.  Thoughts?

lexelby:
Fixed by #201

PR #201, opened 2017-01-17
open prefilled gmail compose window

lexelby:
Partially implements #179 

Instead of sending emails directly from retrodot, this opens a new tab with gmail.  The retro email is already filled out and ready for you to add email addresses.  A couple of downsides:

* Special formatting isn't allowed, so we can't bold/red the "postmortem required" note.
** I used unicode :warning: characters instead.
* Retrodot won't know whether the email has been sent.
** We could set up a special gmail account and use their webhook support to notify retrodot.  Then we'd CC that account on all retro emails.
** Alternatively, we could just assume that if they click "Send retro email" then they follow through and send it.  We'd add an onclick handler to do an Ajax-style request to tell retrodot that they clicked the link.

cc @joshuatobin @jmervine @reidmix


app/admin/incident.rb
```diff
@@ -141,17 +141,24 @@ def prepare_document_links
...
+
+          Retrospective Notes Doc: #{incident.google_doc_url}
+
+          #{"⚠ This incident will need a public postmortem by #{incident.followup_on.to_date}. ⚠" if incident.review?}
+
+          Thanks,
+          SRE
+        EOF
+
+        item "Send Retro Email", "https://mail.google.com/a/heroku.com/mail?view=cm&fs=1&cc=#{cc}&su=#{subject}&body=#{body}", target: "_blank"
```
joshuatobin: Do we really want to hard code this in?

lexelby: Haha, nope.  Totally forgot to fix that, thanks.

jmervine: Wondering if we should consider pull the email contents in to a haml or erb file that's rendered instead of including it in the code. Might make for easier modification later. Thoughts?

lexelby: I considered that.  I just have no idea how to do it, and the rails documentation was confusing.  Any tips?

I don't think HAML's the way to go since I think that's HTML-specific.  Not sure though.

jmervine: I don't have any tips for the Railsy, I would probably go down the route of calling the Haml around parser manually.

HAML is designed with HTML in mind, but can be used for this and I would stick with a single templating language for consistency.

That said, this is probably good for the purposes of this PR if there isn't a quick solution, but I'd be interested in @reidmix's thoughts and possible tips.

joshuatobin:
I'd vote to just :shipit: for now. I don't see us using ActiveAdmin long term. One thing I'd like to see though is having us setup a retrospectives@ list, and then just forcing the CC to send to that list. That way we have some type of tracking until we refactor all of this.

joshuatobin:
👍

jmervine:
👍

lexelby:
After a discussion with @reidmix, I added a commit that marks the retro email sent when they click the link.  That way we don't lose the "NOT SENT" / "DELIVERED" functionality in this PR.  Later on, we can set up an email box for retrodot and have it add itself to the CC list to detect when the email is actually sent.

@joshuatobin @jmervine Can you please give the latest commit a once-over?  My approach is a bit different now, using `window.open` for the gmail compose window, because it simplified the code.

app/admin/incident.rb
```diff
@@ -141,17 +135,30 @@ def prepare_document_links
...
-          "data-subject" => "Incident \##{incident.incident_id} retrospective needed",
-          "data-inputs"  =>
-            {
-              To:      :text,
-              CC:      :text,
-              Subject: :text
-            }.to_json
+        cc = Rack::Utils.escape(Config.email_cc)
+        subject = Rack::Utils.escape("Incident \##{incident.incident_id} retrospective needed")
+        body = Rack::Utils.escape(<<~EOF)
```
reidmix: may be able to put these in the `locales/en.yml` file which will keep the email content out of the code.

lexelby:
I switched to using I18n as per @reidmix's suggestion.  I like how it turned out!

PR #208, opened 2017-01-25
Refactor webhooks (again)

joshuatobin:
There are a bunch of edge cases in this endpoint. I've tried to do the following thus far:

* Be prescriptive in the `update_type` we're willing to accept
* Try to limit the number of fetches based upon the `update_type` and exit as early as possible. 

Remaining edge cases:
- [x] retroactive update_types not properly resolving

A retroactive incident will fire off 2 webhooks, one with the `update_type`=`retroactive` and one for `resolved`. 

This seems to do the right thing, it sees there is not incident present, fires off a sync, and then starts the incident:
```
app=retrodot-staging request_id=7d6942bd-be42-48e5-993e-96e3d18426e1,7760f991-742a-e865-407c-6a036c23eb92 at=sync incident_id=1371 update_type=retroactive message="Recorded the start of chat for incident `1371` at *2017-01-25 11:50:37 -0800*.
```
We then see the in the logs the `resolved` update comes through:

```
Parameters: {"id"=>4010, "created_at"=>"2017-01-25T14:49:55.000Z", "updated_at"=>"2017-01-25T19:50:36.064Z", "incident_id"=>1371, "contents"=>"resolved", "title"=>"retroactive->resolved", "update_type"=>"resolved", "systems"=>[], "webhook"=>{"id"=>4010, "created_at"=>"2017-01-25T14:49:55.000Z", "updated_at"=>"2017-01-25T19:50:36.064Z", "incident_id"=>1371, "contents"=>"resolved", "title"=>"retroactive-resolved", "update_type"=>"resolved", "systems"=>]}}
```
```
app=retrodot-staging request_id=4f7c47e6-ee22-4320-bd27-e235102cdbc3,dc189c9b-b203-5420-e1dd-0fe054f1cb1f at=sync incident_id=1371 update_type=resolved message="Recorded the end of chat for incident `1371` at *2017-01-25 11:50:37 -0800*."
```
However the incident state remains "open"

- [x] Investigating -> false-alarm

Both investigating and false-alarms `update_type` will never send webhooks because they have a `silent` attribute in status, however upon a sync either by cron or when the next incident starts, the incident will be left dangling in an open state. 

- [x] scheduled -> false-alarm
If we go from schedule to false-alarm, again the incident will be left dangling in an open state. 

@jmervine @lexelby @reidmix


joshuatobin:
One thing that came to mind was we could auto-resolve retroactive incidents when we receive the update_type, that way we don't have to rely on ordering.

joshuatobin:
Second idea: Make the persister mediator resolve incidents. If resolved_at date is true set the state to "resolved" blah.

lexelby:
The issue with false-alarm incidents sticking around is due to #216.

lexelby:
I decided to have the webhook endpoint ignore retroactive incidents entirely.  By the time I finished getting everything working, the result was that a retroactive incident caused retrodot to start and then immediately end the timeline.  This doesn't seem particularly useful to me.  If the user wanted a timeline, they probably already started one before the retroactive incident was published.

lexelby:
@reidmix @jmervine I rejiggered this PR quite a bit, and now it's ready for another review.

lexelby:
investigating should also start the timeline

app/controllers/api/v1/webhooks_controller.rb
```diff
@@ -1,85 +1,73 @@
...
 
-  def post_to_slack?
-    start_incident? || resolved? || maintenance?
-  end
-
-  def sync_incident
-    Mediators::Incident::OneSyncher.run(id: incident_id)
+  def check_update_type
+    head :no_content unless UPDATE_TYPES.include? update_type
   end
```
reidmix: You may want to [halt the callback chain](http://weblog.rubyonrails.org/2015/1/10/This-week-in-Rails/#halting-callback-chains-by-throwing-abort) in the case of unsupported `update_type`.  That way it won't execute any of the actions.

I'm thinking that might help you clean up some logic in `sync`?

app/controllers/api/v1/webhooks_controller.rb
```diff
@@ -1,85 +1,73 @@
...
-    check_params[:update_type] == "maintenance"
+  def fetch_incident
+    @incident = Incident.find_by(incident_id: incident_id)
   end
 
   def end_incident
-    ChatOps::Commands::EndCommand.process(User.system_user, "end #{incident_id}")
+    command = "end #{incident_id}"
+    log(at: :sync, command: command)
+    ChatOps::Commands::EndCommand.process(User.system_user, command)
```
reidmix: This may be for a later discussion: but seems to me there would be a way to invoke a command without having to construct the text of the command.  Maybe something like:
```ruby
ChatOps::Commands::EndCommand.run(user)
ChatOps::Commands::StartCommand.run(user, incident_id: 834)
```

app/controllers/api/v1/webhooks_controller.rb
```diff
@@ -1,85 +1,73 @@
...
+      if update_type == "resolved"
+        end_incident[:message] if @incident.timeline?
+      elsif !@incident.timeline?
         start_incident[:message]
-      elsif resolved?
-        end_incident[:message]
-      else
-        "Ignoring unknown update_type: '#{update_type}'."
       end
   end
```
reidmix: This method hurts my head to parse 😄  I'm sure it has to do with my unfamiliarity with it.

reidmix: I did this to get my head around it (which is not much different from the diff):
```ruby
    if @incident.timeline?
      end_incident[:message] if resolved?
    else
      start_incident[:message]
    end
```

But I think it's flawed by missing a case, maybe you can confirm for me: is there a situation where an `update_type` is "resolved" and has no timeline, thus we wouldn't want to `start_incident`?  Perhaps a retroactive incident?

app/models/incident.rb
```diff
@@ -40,8 +40,8 @@ def retro_prepared?
     !trello_url.nil? && !google_doc_url.nil?
   end
 
-  def missing_timeline?
-    timeline_start.nil?
+  def timeline?
+    !timeline_start.nil?
```
reidmix: Also: `timeline_start.present?`

app/controllers/api/v1/webhooks_controller.rb
```diff
@@ -1,85 +1,73 @@
...
+    return render json: t("slack.errors.webhook.incident"), status: 500 unless @incident
 
     process_webhook
 
-    post_to_slack(@message, Config.slack_bot_channel) if post_to_slack?
-    log(at: :sync, incident_id: incident_id, update_type: update_type, message: @message)
+    if @message
+      log(at: :sync, incident_id: incident_id, update_type: update_type, message: @message)
+      post_to_slack(@message, Config.slack_bot_channel)
+    end
```
lexelby: It's expected.  Example: IC already started an incident, and then the status site posting was created.  We don't want retrodot to try to re-start the incident (it'll just throw a spurious error on the slack channel).

app/controllers/api/v1/webhooks_controller.rb
lexelby: As written, this halts the callback chain already.  See section 8: http://guides.rubyonrails.org/action_controller_overview.html

```ruby
redirect_to new_login_url # halts request cycle
```

app/controllers/api/v1/webhooks_controller.rb
lexelby: >  is there a situation where an update_type is "resolved" and has no timeline

Yes.  Example: retrodot misses the first webhook with type "incident", because it was down.  Status will retry a few times but it does give up.  Or perhaps status has a bug and doesn't send the "incident" webhook.

In the usual case, retrodot does "start incident" when receiving a webhook or the human starts the incident manually.  Just starting an incident adds a timeline entry, so the timeline will exist.

Retroactive incidents are a good point.  Status models them as an incident with two updates: the first update is of type "retroactive" and is the description of what happened ("On ___, ___ was broken") and the second is of type "resolved".  I should add "retroactive" to `UPDATE_TYPES` because otherwise, when retrodot tries to end the incident, I think it might spew a spurious "not started" error to chat.  I'll test that.

app/controllers/api/v1/webhooks_controller.rb
lexelby: Interesting idea.  I'll think on how to implement it.

app/models/incident.rb
lexelby: Ah, nice!

app/controllers/api/v1/webhooks_controller.rb
lexelby: Ah, I see I was wrong, and I must have coded with the above cases in mind.  Retrodot ignores the first "retroactive" post, and then ignores the "resolved" because `@incident.timeline?` returns false.  That means a timeline isn't started automatically at ALL by retrodot, and there's no mention in slack from retrodot about a retroactive.

However, if the humans do `/i start` BEFORE posting a retroactive to the status site, then retrodot will end it once we post.  That seems mostly reasonable.

joshuatobin:
I didn't realize this was still a thing. I'll continue to review, thanks @lexelby for picking this up!

app/controllers/api/v1/webhooks_controller.rb
```diff
@@ -1,85 +1,73 @@
...
     @message =
-      if maintenance? || start_incident?
+      if update_type == "resolved"
+        end_incident[:message] if @incident.timeline?
+      elsif !@incident.timeline?
         start_incident[:message]
-      elsif resolved?
-        end_incident[:message]
-      else
-        "Ignoring unknown update_type: '#{update_type}'."
```
joshuatobin: I still think you should be logging as much as possible here. The reasoning for some of this logic is that it was a totally pain in the ass to troubleshoot.

lexelby:
@joshuatobin Sure thing!  I run into the webhook edge-cases fairly often, so I definitely want to get this thing rolled out.

app/controllers/api/v1/webhooks_controller.rb
lexelby: Good idea, will do.

lexelby: Done in 3ab51ee.

I think I came very very close to angering the AbcSize beast. ;)

lexelby:
Added some specs to test all the weird cases, and also some logging per @joshuatobin's suggestion.  @reidmix you've still got a "requested changes" review.  Could you please give this another once-over?

reidmix:
I don't even remember!  I'll review.

app/controllers/api/v1/webhooks_controller.rb
```diff
@@ -1,85 +1,86 @@
...
-  def post_to_slack?
-    start_incident? || resolved? || maintenance?
-  end
-
-  def sync_incident
-    Mediators::Incident::OneSyncher.run(id: incident_id)
+  def check_update_type
+    ignore = !UPDATE_TYPES.include?(update_type)
+    log(at: :check_update_type, ignore: ignore)
+    head :no_content if ignore
```
reidmix: Is the intent for this to halt the callback chain?

http://blog.bigbinary.com/2016/02/13/rails-5-does-not-halt-callback-chain-when-false-is-returned.html

lexelby: I'm not depending on `false` to halt the callback chain.  `head` does: https://github.com/heroku/retrodot/pull/208#discussion_r109026318

app/controllers/api/v1/webhooks_controller.rb
```diff
@@ -1,85 +1,86 @@
...
+    fetch_incident
+    return render json: t("slack.errors.webhook.incident"), status: 500 unless @incident
 
-    process_webhook
+    @message = process_webhook
 
-    post_to_slack(@message, Config.slack_bot_channel) if post_to_slack?
-    log(at: :sync, incident_id: incident_id, update_type: update_type, message: @message)
+    if @message
+      log(at: :sync, incident_id: incident_id, update_type: update_type, message: @message)
```
lexelby: fixed in f769bdd

lexelby:
@reidmix okay, hopefully that should be it.

app/controllers/api/v1/webhooks_controller.rb
reidmix: TIL

PR #209, opened 2017-01-25
Seed a system user 

joshuatobin:
This is a quick hack for adding a system user for webhook commands. Right now we're adding timeline entries via StartCommand and EndCommand, which are returning `Unknown` for a user id, which results in the `CreateRetroDoc` automation failing. **/Note that running Prepare Retro is currently broken in production/**

Feel free to merge this, change it, or close it if I'm out for a bit :)


jmervine:
👍

PR #210, opened 2017-01-27
support missing responder in timeline entries

jmervine:
I've updated the google script to not fail when a responder is missing from a timeline entry. While we are working on handling this via retrodot -- injecting a default user -- this should make the google script a little more solid.

@lex It doesn't look like I have the ability to deploy this script. I'm getting and error -- "Failed to retrieve script's execution endpoint information." -- and best I can tell, it's due to permission issues.


jmervine:
Note: This is just an added (and hopefully unnecessary) layer of error handling. It isn't really critical, but is the safer change.

TBH: I'd prefer that the script succeed should any part of the data be missing. A partial template is better than no template at all, especially when any break fixes would most likely require a code change and release.

joshuatobin:
👍

lexelby:
Yup, it was a permissions issue.  For whatever reason, a Google Script can't be deployed as an API executable by anyone but the owner, regardless of the script's permissions.  I've deployed the new version.

jmervine:
Sweet! Thanks!

Issue #211, opened 2017-01-30
Hide system user from `/i who`


jmervine:
Reference...

```
Retrodot BOT Incident responders for incident `1041`:
- D⁣o⁣u⁣g McInnes <@⁣d⁣o⁣u⁣g>
- K⁣a⁣r⁣a Louie <@⁣k⁣a⁣r⁣a>
- Troels Thomsen <@⁣t⁣t>
- Retrodot <@>
```

joshuatobin:
Whoops. Didn't even think of that one :/

PR #212, opened 2017-01-31
ignoring system user from /i responders

jmervine:
addressing #211


spec/lib/chat_ops/message_helpers_spec.rb
```diff
@@ -0,0 +1,28 @@
+RSpec.describe ChatOps::MessageHelpers do
+  describe ".return_responders" do
+    let(:userR) { create(:retrodot_user) }
+
+    let(:incident) do
+      i = create(:incident_with_responder)
+      i.responders << userR
+      i.save!
+      i
+    end
```
joshuatobin: I believe you can do stuff like this in the factory with after create. This way is fine but I thought i'd mention.

joshuatobin:
👍

spec/lib/chat_ops/message_helpers_spec.rb
jmervine: Maybe but I need you to teach me that crazy Gandalf style magic. I couldn't figure it out. 😂

joshuatobin: Yeah its weird. :shipit: for now

Issue #185, opened 2017-01-10
35 Incident cards for 1009


reidmix:
Not sure what happened to incident 1009 but about +/- 40 cards got created in trello for it.  Thankfully only one gdoc got created for all of them.   I'm putting them into a "Triage" column for investigations.

joshuatobin:
🙀

lexelby:
Maybe my fault?  Not sure?  I thought I did my testing in staging but perhaps I did some in production.

joshuatobin:
Going to label this as wont fix. Looks like testing was done in the wrong env.

lexelby:
Turns out it was due to a redirect loop: #214

PR #222, opened 2017-02-10
Fix "prepare retro" redirect loop

### LINKED ISSUE #214: redirect loop in "prepare retro"
Any error in the creation of the trello card or google doc is treated as an authentication error, and the user is redirected through the OAuth flow.  If the problem is not actually authentication-based (e.g. if it's a matter of permissions/authorization), then this results in a redirect loop.

Similarly, if a user's Google login session has expired, in some cases, Google OAuth will silently return failure to our oauth callback.  This also results in a redirect loop.

lexelby:
This fixes #214 (redirect loop) and #185 (many trello cards created).  It also adds a "loading" spinner animation for the user to watch while "prepare retro" does its thing.

@reidmix @jmervine


Procfile
```diff
@@ -1,2 +1 @@
 web:     bundle exec puma -t ${MIN_THREADS:-2}:${MAX_THREADS:-6} -w ${WEB_WORKERS:-2} -p ${PORT:-3000} -e ${RACK_ENV:-development}
-release: bundle exec rake db:migrate db:seed
```
jmervine: Why remove this? It can be moved in to `app.json` as well.

app/admin/incident.rb
```diff
@@ -61,17 +61,25 @@ def prepare_document_links
...
+      else
+        session.delete(:redirects)
+        redirect_to collection_path, flash: { error: "Auth failure: #{e}" }
+      end
+    rescue => e
+      session.delete(:redirects)
+      log(error: e, at: "prepare retro")
+      redirect_to collection_path, flash: { error: "Prepare Retro failed: #{e}" }
     else
+      session.delete(:redirects)
```
reidmix: Looks like `session.delete(:redirects)` occurs in all cases but one -- anyway to handle this just once?

app/admin/incident.rb
```diff
@@ -61,17 +61,25 @@ def prepare_document_links
...
-      session[:return_to] = prepare_retro_admin_incident_path(resource[:id])
-      redirect_to "/auth/google_oauth2"
-    rescue NoMethodError => e
-      redirect_to collection_path, flash: { error: "Error: #{e}" }
+    rescue Exceptions::RedirectRequired => e
+      session[:redirects] ||= 0
+      session[:redirects] += 1
+
+      if session[:redirects] <= 2
+        session[:return_to] = prepare_retro_admin_incident_path(resource[:id])
```
reidmix: `TrelloAuthRequired` didn't originally call `prepare_retro_admin_incident_path` -- is calling it in this case going to have side-effects?

lib/exceptions.rb
```diff
@@ -0,0 +1,19 @@
...
+  # A mediator would like to redirect the user.
+  class RedirectRequired < StandardError
+    attr_accessor :url
+
+    def initialize(message, url)
+      super(message)
+      @url = url
+    end
+  end
+end
```
reidmix: FWIW: Personally, I tend to keep exceptions as close to the class that is emitting them, this keeps them together conceptually.  It also avoids having a bucket where all exceptions go where they may lose their context.  That being said I'm not sure I have a great suggestion on where I'd put these since they're all "wrapping" exceptions from libraries in the mediators.

Procfile
lexelby: Oops, that wasn't supposed to make it here to the PR.  I did that so that I could quickly iterate on staging.  Kind of a hack.

app/admin/incident.rb
lexelby: `prepare_retro_admin_incident_path` is just a URL helper, so no, there won't be any side-effects.  In fact, this control flow _did_ eventually end up calling that url helper in TrelloController.  The logic was kind of confusing.

lib/exceptions.rb
lexelby: I see your point there.  I went with this pattern for two reasons: 1) there was a very highly-rated stack overflow answer suggesting this pattern, and 2) because otherwise we end up having to declare those exceptions in multiple places.  That's how the code used to be, and that's what I'm fixing by moving these here.

app/admin/incident.rb
lexelby: I know -- it bothers me too.  So far, I haven't been able to think of a way to do it that doesn't make it _less_ readable.  It's down to the situations in which I need to clear the session var: on success, on all other kinds of exceptions, or on auth exceptions when we've already retried twice.

Ideas I've had:

1. clear it when viewing the incident list page
2. clear it after the `begin/rescue/else` block and return early in the case where I _don't_ want it cleared

1 is going to seem impromptu, and 2 is going to be subtle and "clever".

lib/exceptions.rb
reidmix: yup, there's not really a logical place for them in this case.  and putting them on a mediator base class feels weird too.  Anyways 👍 😄

app/admin/incident.rb
jmervine: I'd vote  for `2`. It is subtle, but not so much so that it should impact readability.

app/assets/javascripts/active_admin.js
```diff
@@ -1 +1,20 @@
 //= require active_admin/base
+
+$(document).on('ready page:load turbolinks:load', function() {
```
jmervine: `ready page:load turbolinks:load` seems overkill. Are you addressing a specific issue, or just defaulting to that? I'd typically just use a basic closure -- `$(function() { ... my code ... });` -- is why I ask.

lib/exceptions.rb
jmervine: Honestly, I tend to avoid creating  custom exception -- although I know some would strongly disagree -- in favor or raising a `RuntimeError` with a custom message via `raise "my message"`. It avoids exactly what we're discussing here, while providing the same (or similar) basic feedback.

lexelby: In this case I'm actually using a custom field in the exception to pass the path to redirect to back up to the controller code.

I also need to differentiate between a trello auth failure and a google auth failure.  I would suggest that comparing `e.message` is a pretty big anti-pattern as opposed to rescuing a specific custom exception type.  I could avoid this case by wrapping the trello and google auth failure handlers in separate `begin/rescue` blocks, but that would require duplicating the trello auth exception handling because there are three steps: trello, google, trello.

app/assets/javascripts/active_admin.js
lexelby: Honestly, I haven't the faintest clue.  I believe I snitched that from something online ages ago, or perhaps I got it from in the activeadmin code.

What does wrapping a function in `$()` do?

jmervine: `$()` is short for `jQuery()`, which executes once the DOM is ready. See http://api.jquery.com/jquery/#jQuery3.

lexelby: Thanks!  I couldn't figure out how to google for that.

app/admin/incident.rb
reidmix: Agreed, `2` sounds good.

app/support/mediators/incident/create_card.rb
```diff
@@ -29,9 +27,9 @@ def call
         "idCardSource"   => @template.id,
         "idList"         => Config.trello_list_id || @template.list_id,
         "keepFromSource" => "all")
-    rescue Trello::InvalidAccessToken, Trello::Error, NoMethodError
+    rescue Trello::InvalidAccessToken, Trello::Error, NoMethodError => e
       log_error($ERROR_INFO, fn: "call", at: "run", id: @id)
```
jmervine: If you're going to set `e` in the `rescue`, might as well use `e` over `$ERROR_INFO` in `log_error`.

app/support/mediators/incident/create_retro_doc.rb
```diff
@@ -18,9 +15,9 @@ def call
       log(fn: :call, at: :create_retro_doc)
       resp = @service.run_script(Config.google_script_id, execution_req)
       resp.to_h[:response][:result]
-    rescue Google::Apis::ClientError, Google::Apis::AuthorizationError, Signet::AuthorizationError
+    rescue Google::Apis::ClientError, Google::Apis::AuthorizationError, Signet::AuthorizationError => e
       log_error($ERROR_INFO, fn: "call", at: "run", id: @id)
```
jmervine: Same, if setting `e`, use `e` over `$ERROR_INFO`.

app/support/mediators/incident/prepare_retro.rb
```diff
@@ -8,42 +8,52 @@ def initialize(incident:, current_user:)
...
-      @incident.update(trello_url: @trello.url, google_doc_url: @gdoc)
-
-    rescue GoogleAuthRequired
-      log_error($ERROR_INFO, fn: "call", at: "CreateRetroDoc", incident: @incident.id)
-      raise
-    rescue TrelloAuthRequired
-      log_error($ERROR_INFO, fn: "call", at: "CreateCard", incident: @incident.id)
-      raise
+    rescue Exceptions::TrelloAuthRequired => e
+      log_error($ERROR_INFO,
```
jmervine: `s/$ERROR_INFO/e/`

app/support/mediators/incident/prepare_retro.rb
```diff
@@ -8,42 +8,52 @@ def initialize(incident:, current_user:)
...
+                at: :create_trello_card,
+                fn: :member_action)
+      raise Exceptions::RedirectRequired.new(e.message, "/auth/trello")
+    rescue Exceptions::GoogleAuthRequired => e
+      log_error($ERROR_INFO,
```
jmervine: `s/$ERROR_INFO/e/`

app/support/mediators/incident/create_card.rb
reidmix: I so agree with this -- i'm not a fan of `$ERROR_INFO` over `$!`.

jmervine:
👍

PR #223, opened 2017-02-10
react instead of PMing

### LINKED ISSUE #206: Have slack reactions respond w/ a checkmark, not PM the user
CC: @danp

lexelby:
Fixes #206.  Instead of sending a PM to tell someone that retrodot saw them react with the timeline reaction, it just reacts with a checkmark.


lexelby:
Also made it so that we can specify multiple timeline reaction emojis, because people don't like the name of our current one.

lexelby:
@jmervine @reidmix this is ready for review

app/controllers/api/v1/chat_ops_controller.rb
```diff
@@ -61,10 +61,12 @@ def slack_event
     )
 
     current_incident.responders << user
-    entry = [Config.chatops_message_confirmation, timeline_success_msg]
-    post_to_slack(entry.join("\n"), slack_user_from_event)
 
-    render json: t("slack.timeline.added"), status: 200 # Render 201 so slack doesn't retry
+    post_slack_reaction(event_params[:event][:item][:channel],
+                        event_params[:event][:item][:ts],
+                        Config.confirmation_reaction_emoji)
```
reidmix: Looks like there's helpers for some of these (like `slack_channel_from_event`) but not all.

I've looked at this code before and wondered if we should just be passing in just the hash for `event_params[:events]`  or `event_params[:events][:item]` in this case -- maybe as named params -- that might simplify the interface.

app/controllers/api/v1/chat_ops_controller.rb
```diff
@@ -143,7 +140,7 @@ def render_slack_challenge
   def slack_reaction?
     event_params[:event][:type] == "reaction_added" &&
       event_params[:event][:item][:type] == "message" &&
-      event_params[:event][:reaction] == Config.timeline_reaction_emoji
+      Config.timeline_reaction_emoji.include?(event_params[:event][:reaction])
```
reidmix: Sweet!

app/controllers/api/v1/chat_ops_controller.rb
lexelby: I tidied it up.  If you get a chance, could you re-review?

jmervine:
👍

lexelby:
Fixed a couple bugs, tested on staging, and confirmed that now it works.

config/config.rb
```diff
@@ -70,5 +70,6 @@ module Config
   optional :slack_slash_command_token, string
   override :slack_bot_username, "retrodot", string
   override :slack_bot_channel, "retrodot", string
-  override :timeline_reaction_emoji, "timeline_add", string
+  override :timeline_reaction_emoji, "timeline_add", array(string)
```
jmervine: +1

jmervine:
👍 🚢

PR #231, opened 2017-02-16
Edit category, primary_team, etc in incident list

lexelby:
This adds javascript piping to allow the user to edit the following fields of incidents right in the incident list view:

* category
* primary team
* retro email sent (checkbox)
* retro held date

It's written such that we can add more editable columns very easily for boolean, date, and belongs_to fields.  Other kinds of fields would be pretty easy too.

"belongs_to" fields like category and primary team have a dropdown of existing options along with a "Custom..." option.  Choosing "Custom..." lets you enter a new category/team that is created automatically when you press enter.

While I was in there, I added an optimization that speeds up the incident list page load time significantly.  ActiveAdmin was pulling the incident responders list for every single incident in existence (all thousand) to populate a "filter" dropdown on the right, so that you could, for example, search for all incidents that some specific person was in.  Disabling this prevents it from having to load every single incident, and I don't think that kind of filter will be used very often if ever.


lexelby:
Oops, forgot to run rubocop.  Will fix tomorrow.

lexelby:
The PR app is loaded up with incidents, so you can test out the new live-editing functionality now.

lexelby:
Massive rubocop run.  Also moved the editable field stuff into better locations so it can be used in all kinds of resources.  It could probably be gemified, even.

Added tests for `format_duration`, too.

@reidmix @jmervine This is ready for review.

app/helpers/active_admin/views_helper.rb
```diff
@@ -0,0 +1,100 @@
...
+            content_tag(:option,
+                        item,
+                        selected: current_value&.name == item)
+          end
+        ]
+      )
+    end
+  end
+
+  def editable_container(resource, attr, **attrs, &block)
```
reidmix: because `attrs` are the option at the end, you can drop the asterisks, ruby should collect all the key-values for you.

app/assets/javascripts/retrodot.js
```diff
@@ -0,0 +1,217 @@
...
+  $.fn.flash_message = function(options) {
+
+    options = $.extend({
+      text: 'Done',
+      time: 1000,
+      how: 'after',
+      class_name: 'flash_message'
+    }, options);
+
+    return $(this).each(function() {
```
jmervine: I don't **think** the the `return` is necessary here.

app/assets/javascripts/retrodot.js
```diff
@@ -0,0 +1,217 @@
...
+            });
+          }
+        },
+      });
+
+      input.focus();
+    });
+
+    $("div.editable_foreign_key_column input").on("change", function(e) {
+    });
```
jmervine: This appears to be doing nothing.

app/helpers/active_admin/views_helper.rb
```diff
@@ -0,0 +1,100 @@
+# Helper methods for all ActiveAdmin resources, including quick-edit columns.
+module ActiveAdmin::ViewsHelper
+  def format_duration(seconds)
+    seconds = seconds.to_i # in case they pass a Duration object
+    description = convert_seconds_to_description(seconds)
+    strip_zeroes(description)
+  end
+
+  private
```
jmervine: I would move all public methods to the top and private methods under a single `private` call. Reads cleaner.

app/assets/javascripts/retrodot.js
lexelby: TBH, I'm not sure.  I copied this code from some jsfiddle, so I'm just going to leave it as it is.

app/helpers/active_admin/views_helper.rb
lexelby: I did it this way hoping that it would make sense contextually.  `convert_seconds_to_description` and `strip_zeroes` are only used by `format_duration`.  With that in mind, do you still feel like they should be moved down?

lexelby:
@reidmix ready for a re-review

app/admin/incident.rb
```diff
@@ -109,25 +113,34 @@ def prepare_document_links
...
+      l(i.started_at.to_date, format: :retrodot_short)
+    end
+    column :duration do |i|
+      format_duration(i.duration)
+    end
+    column :state do |i|
+      i.state if i.state != "resolved"
+    end
+    column :title do |i|
+      link_to i.title, "#{Config.status_site_url}/#{i.incident_id}", target: "_blank"
```
reidmix: In the future, it might be nice to have that status_url built right from the incident model object.

app/helpers/active_admin/views_helper.rb
reidmix: It's not common to switch back and forth between public and private that I've seen -- for me personally, this is why I tab everything under private because it's hard to see visually that we've moved into a different (modal) state.  

TBH once I saw the `private` keyword I would generally assume the rest was private, I'd suggest it be moved down.

app/helpers/active_admin/views_helper.rb
```diff
@@ -0,0 +1,100 @@
...
+
+  def convert_seconds_to_description(seconds)
+    mm, ss = seconds.divmod(60)
+    hh, mm = mm.divmod(60)
+    dd, hh = hh.divmod(24)
+    "#{dd}d #{hh}h #{mm}m #{ss}s"
+  end
+
+  def strip_zeroes(description)
+    description.gsub(/(^|\s)(0d|0h|0m|0s)/, "").gsub(/\s+/, " ").strip
```
reidmix: I believe could you solve this in `convert_seconds_to_description`:
```ruby
"#{dd}d #{hh}h #{mm}m #{ss}s" if ss.zero? && mm.zero? && hh.zero? && dd.zero?
```

app/helpers/active_admin/views_helper.rb
```diff
@@ -0,0 +1,100 @@
...
+      )
+    end
+  end
+
+  def editable_container(resource, attr, attrs, &block)
+    content_tag(:div,
+                "data-name" => resource.model_name.collection,
+                "data-attr" => attr,
+                "data-resource-id" => resource.id,
+                **attrs,
```
reidmix: even though I see this works, this `**attrs` syntax is so weird to me -- i need to look it up to understand it better.

app/helpers/active_admin/views_helper.rb
lexelby: Got it.  Will do!

app/admin/incident.rb
lexelby: Sure, I'll just throw that in now.

app/helpers/active_admin/views_helper.rb
lexelby: I don't see how that logic solves the problem at all.  I think I see what you're saying though: I could build an array of days, hours, minutes, and seconds and then join with a space.

app/helpers/active_admin/views_helper.rb
lexelby: Yeah, it's a lot more common in python.  Since I'm already passing some args by name, I have to use `**attrs`.  The only other option would be to merge the `data-*` args into `attrs` and then pass `attrs` plain without `**`, but I think that'd be more clunky.

app/helpers/active_admin/views_helper.rb
reidmix: I was thinking that it wouldn't output any message in the case of all zeros, which is getting regexed out later.

app/helpers/active_admin/views_helper.rb
reidmix: It's a nice shortcut 👍 I just hadn't wrapped my head around why it works in ruby :D I'm guessing it's just it's own feature and it's not de-referencing it twice.

app/helpers/active_admin/views_helper.rb
lexelby: Ah, I see.  The regex isn't to catch the case of 0 seconds; it's to remove ALL zero components.  Example:

1d 0h 0m 12s -> 1d 12s

Note that it's using `|` and `gsub` so it'll catch each 0 component separately.  Then it cleans up any extra spaces left behind.  Perhaps the array method I suggested above would be clearer, since it would allow us to get rid of `strip_zeroes` entirely.

app/helpers/active_admin/views_helper.rb
lexelby: Yup. :)  Passing a hash in as keyword args using `**` works the same in python and ruby.  I think I discovered it the first time simply by coding like it was python and hoping it worked ;)

app/admin/incident.rb
lexelby: Done in d1f4527

app/helpers/active_admin/views_helper.rb
lexelby: Nope, the method I suggested above fails rubocop with an AbcSize of 16.12 (threshold is 16.0).  Sigh.

lexelby:
@reidmix okay, ready for one more review :)

app/helpers/active_admin/views_helper.rb
reidmix: Ah of course!  That makes better sense than how I was reading it (incorrectly).

reidmix: If you were to do it, you'd have to do something like:
```ruby
[ "#{dd}d" unless dd.zero?, 
  "#{hh}h" unless hh.zero?,
  "#{mm}m" unless mm.zero?,
  "#{ss}s" unless ss.zero?
].compact.join(" ") 
```

reidmix: We can clean this up later ^ what we have here works and we can at least get this diff out now. 👍

PR #241, opened 2017-03-02
fixing line break in start command

jmervine:
Hehe, didn't notice this until after I merged and tested on staging...

screen shot 2017-03-02 at 8 31 28 am

@lexelby @stellacotton


stellacotton:
👍

PR #242, opened 2017-03-03
Try including the production timeline url

atmos:
This is an attempt to make teams aware of the production timeline in TPS.

At the start of an incident we link to the production timeline so teams responding to the incident can quickly skim what's changed recently across the platform.


stellacotton:
We might want to only add this if that config var exists? It will be weird if you have no timeline.

@jmervine for context, we forgot that this is open source and this is a pretty specific-to-us thing

lib/chat_ops/commands/start_command.rb
```diff
@@ -56,7 +56,7 @@ def run
       incident.save
 
       response << "use `/incident help` for more information on the commands available."
-      response << "\n<https://tps.heroku.tools/timeline|View recent platform changes.>"
+      response << "\n<#{Config.production_timeline_url}|View recent platform changes.>"
```
jmervine: Can we make this...

```ruby
response << "\n<#{Config.production_timeline_url}|View recent platform changes.>" if Config.production_timeline_url?
```

.env.development.sample
```diff
@@ -7,3 +7,4 @@ RESOLVED_AT=resolved_at
 REVIEW=review
 FOLLOWUP_DAYS=5
 LOG_LEVEL=debug
+PRODUCTION_TIMELINE_URL="https://tps.dev/timeline"
```
jmervine: Because this is OOS, I would remove mention of `tps` all together, perhaps just `https://timeline.dev/timeline`.

.env.test
```diff
@@ -10,3 +10,4 @@ FOLLOWUP_DAYS=5
 CHATOPS_API_KEY=sekrit
 EXPORT_API_KEY=sekrit
 LOG_LEVEL=debug
+PRODUCTION_TIMELINE_URL="https://tps.reports/timeline"
```
jmervine: As mentioned above...

Because this is OOS, I would remove mention of `tps` all together, perhaps just `https://timeline.test/timeline`.

jmervine:
@atmos (cc @stellacotton ) Perfect! 👍

PR #257, opened 2017-04-20
fix slash command errors

### LINKED ISSUE #227: Error when using /i help
retrodot-error

When using `/i help`, I get both the help text and an error message- see attachment.

lexelby:
Typo, should fix #227.  I also removed the original command string from retrodot's response, since it's redundant and apparently unnecessary.  Slack doesn't complain about the empty message field.


joshuatobin:
@lexelby Looks good but there is a failing spec

lexelby:
wtf how did I miss that?

lexelby:
@tobin Fixed.  Apparently I wrote the test to match the incorrect code.  Ephemeral responses shouldn't do the delayed job / POST thing because we don't bother echoing their command to them.  I fixed the specs to expect a direct HTTP response.

joshuatobin:
👍

PR #258, opened 2017-04-20
don't show 'forgot to run start incident' for read-only actions

### LINKED ISSUE #236: "did you really mean incident X" thing is annoying
The "It loos like you may have forgotten to run `start` for this incident." message is actually pretty annoying.  Almost always, I mean to be acting on the most recent incident.  I end up having to copy and paste my command in.  Retrodot could perhaps handle this better, by returning an error to slack, whcih causes one's client to keep the command in the edit line.

lexelby:
Fixes #236 

I adjusted the "It looks like you forgot to run start incident" checking.  Now, it only says that if you're trying to modify something.  For read-only things like `/i timeline` or `/i who`, it just responds.

My thinking is that if the user really did get the wrong incident, they'll see that retrodot returned info for the wrong incident and adjust accordingly.  If they had the right incident in mind, then retrodot has not bothered them incorrectly.


joshuatobin:
@lexelby 👍 just need to fix the spec

joshuatobin:
👍

PR #260, opened 2017-04-21
view timeline in web UI, delete entries

lexelby:
I love how easy activeadmin makes this!  This code lets you click an icon next to an incident to view its timeline.  You can "delete" entries from the timeline -- which really just sets a `deleted` flag to true.  Deleted entries aren't shown via chatops and aren't added to the retro doc.  You can view deleted entries by changing the filter on the timeline entry list page, and you can undelete entries.

I added font-awesome (which was also easy) because the action links were getting a bit crazy. :tada: 

cc @joshuatobin


lexelby:
@joshuatobin ready for review

lexelby:
@heroku/sr-tools Could someone please review this?

jmervine:
This is awesome and a need!

It'd be great if entries could be edited as well, but if that's out of scope for this, then...

:+1:

lexelby:
I'm gonna declare it out-of-scope if only so that I can just get this out there.  It'd be a very easy change to allow editing of these entries using the normal activeadmin "edit row" dialog, and a moderate change to allow editing entries "live" on the collection page like I did for team/category.

PR #261, opened 2017-04-21
timeline batch actions

lexelby:
Adds batch actions to the timeline view:
  * delete
  * undelete (restore)
  * move to another incident

This is branched off of #260.


joshuatobin:
👍

PR #270, opened 2017-05-25
update auth

jmervine:
Replacing existing auth w/ drydock's auth. 

**Why?** Authorization is broken (specifically the refresh token handling). Users frequently have to deauthorize and reauthorize the app. We've fixed this in Drydock and are taking our learning to Retrodot.

**Updates:**

- Added user sync from component inventory.
- ~~Removed user sync from slack.~~
- Updated configerator to support namespaces.
- Removed release command to speed up deployments. **`rake db:migrate` must now be manually run.**
  - Note: this is a pattern we're moving to with all apps, currently.
- Added user to activeadmin.
- Requiring user to update slack handle when not present on first login.
- Replaced auth workflow with workflow from drydock.
- Cleaned up trello auth workflow.
- Renamed User `google_auth_code` to `google_access_token`.
- Re-typed User `email` and `name` to strings.
- Added error handling for edge case redirection loop, with error page (taken from drydock)
  - see: https://github.com/heroku/drydock/issues/146


app/controllers/api/v1/chat_ops_controller.rb
```diff
@@ -26,8 +26,10 @@ def respond
   def slack_slash_command
     return render_slack_error(t("slack.errors.token"), 401) unless valid_slack_token?
 
-    user = User.find_by(slack_user_id: chat_params[:user_id])
-    return render_slack_error(t("slack.errors.user.id"), 403) unless user
+    user = User.where(slack_user_id: chat_params[:user_id])
+               .or(User.where(handle: chat_params[:user_name])).first
+
+    return render_slack_error(t("slack.errors.user.id", Config.frontend_host), 403) unless user.present?
```
jmervine: @joshuatobin Oddly, this check never throws an error, even when I unset slack_user_id and slack_handle on staging. Not sure what's up with that? I was considering removing the user check all together anyway, as it seems overkill with the API token check. Thoughts?

jmervine:
Grrrr....  from slack:

---

Okay, to properly fix authentication in Retrodot, I need to sync users from Component Inventory. The issue is, Retrodot also relies on having the correct slack information. There are ~40 Slack users who created their slack accounts using alias email addresses and therefore there slack data isn’t populated in Component Inventory. Those users will not be able to use Retrodot unless they change their email addresses in Slack to match their primary heroku email address. When this error occurs, I’m having Retrodot send them the following message:

> Retrodot does not recognize your Slack user. This is most likey because the email address you used to register with Slack is an alias to your primary Heorku email address. To use Retrodot, you must update your email address setting in Slack to use your primary Heroku email address. Once updated, it takes two to four hours to propagate. You can find direction on updating your Slack email address here: https://get.slack.help/hc/en-us/articles/207262907-Change-your-email-address.

If I try to add someone to the responders whose slack information is missing, the following error will be thrown:

> Person not found.  Try using their @handle.

Bah, this is a mess.

I can overlay slack users, there will end up bring duplicate users, but it should fix the issue.

---

Which is what I did. I left the error messaging the same, but hopefully they'll never actually be hit.

jmervine:
@sheax0r posted this is slack:

> Google API - How do I get alias emails for a user?
> https://stackoverflow.com/questions/22671064/google-api-how-do-i-get-alias-emails-for-a-user

config/locales/en.yml
```diff
@@ -43,8 +43,8 @@ en:
       webhook:
         incident: "Error starting incident"
       user:
-        id: "Slack `user_id` not found"
-      token: "Invalid Token"
+        id: "Retrodot does not recognize your Slack user. This is most likey because the email address you used to register with Slack is an alias to your primary Heorku email address. To use Retrodot, you must update your email address setting in Slack to use your primary Heroku email address. Once updated, it takes two to four hours to propagate. You can find direction on updating your Slack email address here: https://get.slack.help/hc/en-us/articles/207262907-Change-your-email-address."
```
sheax0r: "Heorku" :)

db/migrate/20170523225609_rename_google_access_token_column_on_users.rb
```diff
@@ -0,0 +1,21 @@
...
+
+    rename_column :users, :google_auth_code, :google_access_token
+
+    users.each do |user|
+      User.find(user.first).update(google_access_token: user.last)
+    end
+  end
+
+  def down
+    users = User.pluck(:id, :google_access_token)
```
sheax0r: I find that it can cause problems somewhere down the line if I use model objects in db migration scripts. eg, what if I end up renaming a class entirely? Old db migration scripts can actually stop working in this case. I have shot myself in the foot this way in the past.

So I'd suggest not using the User class (or any model class) in your migrations. Up to you though; the risk may be small, and something you're willing to deal with down the line if it ever happens.

app/controllers/api/v1/chat_ops_controller.rb
```diff
@@ -179,15 +193,15 @@ def find_slack_user
...
   end
 
   def slack_user_from_event
-    event_params[:event][:user]
+    event_params.dig(:event, :user)
   end
 
   def timestamp_from_event
-    event_params[:event][:item][:ts]
+    event_params.dig(:event, :item, :ts)
```
joshuatobin: 👍 I did all this before I knew about dig :p

app/controllers/application_controller.rb
```diff
@@ -3,14 +3,39 @@ class ApplicationController < ActionController::Base
   protect_from_forgery with: :exception
 
   def current_user
-    return unless session[:user]
-    log(fn: "current_user", at: "find_user", email: session[:user][:email])
-    # ActiveAdmin looks up the wrong namespace for User, so we have to prepend with ::
-    @user ||= ::User.find_by(email: session[:user][:email])
+    return unless session["user"]
+
+    @current_user ||= User.find_by(email: session["user"]["email"])
```
joshuatobin: I remember specifically running into all sorts of issues with the namespace here. Did you not run into this?

joshuatobin: Also 👍 on removing the log here.

Gemfile
```diff
@@ -61,6 +61,5 @@ group :test do
 end
 
 group :development do
-  gem "spring"
-  gem "spring-watcher-listen", "~> 2.0.0"
+  gem "listen"
```
joshuatobin: Just curious what you used this for?

Procfile
```diff
@@ -1,2 +1 @@
-web:     bundle exec puma -t ${MIN_THREADS:-2}:${MAX_THREADS:-6} -w ${WEB_WORKERS:-2} -p ${PORT:-3000} -e ${RACK_ENV:-development}
-release: bundle exec rake db:migrate db:seed
```
joshuatobin: +1

app/controllers/api/v1/chat_ops_controller.rb
```diff
@@ -41,6 +47,7 @@ def slack_slash_command
     render json: { response_type: @response[:response_type] }, status: 200
   end
 
+  # rubocop:disable Metrics/CyclomaticComplexity
```
joshuatobin: I think its easier said then done. I remember wrestling with this a bunch. There are a whole bunch of corner cases which create complexity. I believe previously I tried refactoring a bunch of actions to `before_*` filters but it ended up muddy'n the waters.

app/models/authentication.rb
```diff
@@ -0,0 +1,103 @@
...
+  def login_and_create_user
+    @user ||= \
+      if session["user"].present?
+        User.find_by(email: session["user"]["email"])
+      elsif developer? || valid_email?
+        User.ensure(email: omniauth.email, name: omniauth.name)
+      end
+  end
+
+  def update_missing_user_tokens
```
joshuatobin: Its possible I'm missing something but isn't this going to cause a scenario where sometimes the trello token/secret get `nil`'d out and other times they won't?

app/models/authentication.rb
```diff
@@ -0,0 +1,103 @@
...
+  def bad_refresh_token_state?
+    !developer? && user.present? && !omniauth.refresh_token? && !user.google_refresh_token?
+  end
+
+  def provider
+    self.class.provider
+  end
+
+  def developer?
+    provider == :developer
```
joshuatobin: @sheax0r good point. We might be able to move some of this logic directly into `Config`.

app/support/mediators/incident/persister.rb
```diff
@@ -8,16 +8,17 @@ def initialize(incident:)
     end
 
     def call
-      log(fn: "call")
       update_incident
     rescue Excon::Error
-      log_error($ERROR_INFO, fn: "call", at: "run", incident_data: @incident_data)
+      log_error($ERROR_INFO, fn: :call, incident_data: @incident_data)
```
joshuatobin: I believe that was because rubocop complains.

lib/tasks/sync_slack_users.rake
```diff
@@ -1,6 +0,0 @@
-namespace :sync do
-  desc "Sync users from slack"
-  task slack_users: :environment do |_t|
-    Mediators::User::SyncSlackUsers.run(token: User.with_slack_token.first&.slack_access_token)
-  end
-end
```
joshuatobin: Just a note I believe there is a heroku scheduler job that will need to be deleted / updated.

app/models/component_inventory/user.rb
```diff
@@ -0,0 +1,13 @@
+module ComponentInventory
+  # Map component inventory user api endpoint
+  class User < ComponentInventory::Base
+    attr_accessor :name, :heroku_email, :slack_handle, :slack_user_id
+
+    def self.all
+      fetch(:users).map do |user|
+        new(name: user["name"], heroku_email: user["heroku_email"], slack_handle: user["slack_handle"],
```
joshuatobin: dumb question. Calling `new()` here, is it the same as calling `User.new`?

joshuatobin: Nevermind. I see its completely separate and you're mapping over it later and passing it to User.ensure. 🤕

app/controllers/api/v1/chat_ops_controller.rb
jmervine: I do agree and I plan on doing some systematic refactors to this code base. At this point though, the PR is already so large, I'm trying to keep the scope of changes as small as possible.

app/models/authentication.rb
jmervine: Yeah, currently this is the pattern that we use in all of our apps. It could use some revisiting. I'd rather keep them consistent for the time being and then perhaps take an afternoon and refactor them all in one go.

app/support/mediators/incident/persister.rb
jmervine: Rubocop complains between `$?` and `$ERROR_INFO`. Otherwise, I don't think it makes a difference. I personally prefer to use the more explicit `rescue Excon::Error => e`, again, in the case of this PR I was just trying to change as little as possible.

db/migrate/20170523225609_rename_google_access_token_column_on_users.rb
jmervine: Yes, and I very much agree with you, generally. I in this case, to avoid breaking authentication for those already authenticated, I need to do this. I'll wrap these in rescue blocks though to prevent breakage down the line.

app/controllers/api/v1/chat_ops_controller.rb
jmervine: Hehe, did this out of lazyness, kept me from having to build the full hash in specs.

app/controllers/application_controller.rb
jmervine: I remember that too. I think we fixed that in other ways though in Drydock and I believe it will carry over here.

app/models/authentication.rb
jmervine: Oh, maybe. Or perhaps they'll never get set. I'll need to do some more testing around this.

app/models/component_inventory/user.rb
jmervine: Yeah, `new` is actually `ComponentInventory::User.new` ;)

config/locales/en.yml
jmervine: 🤦‍♂️

lib/tasks/sync_slack_users.rake
jmervine: The mediator is now being called from the other sync rake task `lib/tasks/sync.rake`.

app/admin/user.rb
```diff
@@ -0,0 +1,47 @@
...
+    actions
+  end
+
+  controller do
+    def update
+      user = User.find(params[:id])
+      user.update!(handle: params[:user][:handle])
+
+      if session[:redirect_on_success]
+        redirect_to(session[:redirect_on_success])
```
lexelby: Do you need to remove `:redirect_on_success` from the session?

app/models/authentication.rb
```diff
@@ -0,0 +1,103 @@
...
+  end
+
+  def valid_email?
+    omniauth.email =~ /@#{Config.google_domain}/
```
lexelby: This is worth anchoring at the end of the string with a `$`.

app/models/authentication.rb
```diff
@@ -0,0 +1,103 @@
...
+  end
+
+  def invalid_email?
+    omniauth.email !~ /@#{Config.google_domain}/
```
lexelby: how about `!valid_email?`

app/models/component_inventory/base.rb
```diff
@@ -0,0 +1,18 @@
+module ComponentInventory
+  # Base class for component inventory ingestion
+  class Base
+    include ActiveModel::Model
+
+    def self.fetch(target)
+      return [] unless Config.component_api?
+
+      Rails.cache.fetch("component-inventory-#{target}", expires: 60.minutes) do
+        JSON.parse(Excon.get(
```
lexelby: Lots of stuff could go wrong here -- HTTP request failure, JSON parse failure, etc.  What happens if an exception is raised?

app/models/component_inventory/user.rb
lexelby: Can you use the double-splat (`**`) operator to pass in all those key/value params at once?

config/locales/en.yml
lexelby: Can you use one of YAML's block text things to avoid this super-long line, like `|`?

db/migrate/20170523225609_rename_google_access_token_column_on_users.rb
```diff
@@ -0,0 +1,21 @@
+class RenameGoogleAccessTokenColumnOnUsers < ActiveRecord::Migration[5.0]
+  def up
+    users = User.pluck(:id, :google_auth_code)
+
+    rename_column :users, :google_auth_code, :google_access_token
+
+    users.each do |user|
+      User.find(user.first).update(google_access_token: user.last)
```
lexelby: `user.first` here threw me.  Can you unpack `users` items like this?:

`users.each do |id, token|`

lib/tasks/sync.rake
```diff
@@ -0,0 +1,29 @@
...
+
+  def verify_component_configs
+    valid = Config.component_api?
+
+    puts "  WARNING: Component Inventory API Username (COMPONENT_API_USERNAME) isn't configured." unless Config.component_api_username?
+    puts "  WARNING: Component Inventory API Password (COMPONENT_API_PASSWORD) isn't configured." unless Config.component_api_password?
+
+    valid
+  end
+end
```
joshuatobin: Tested all these in staging and they work as expected 👍

db/migrate/20170523225609_rename_google_access_token_column_on_users.rb
lexelby: I don't quite understand this.  Why do you need to fetch the data before the rename and set it under the new column name after?  Shouldn't the rename of the column leave the values in place?  I think I'm missing an activerecord subtlety here...

config/locales/en.yml
jmervine: It adds new line breaks in the messing. That's how I implemented it originally and then reverted it.

Gemfile
jmervine: I'm not sure, it's used by rails in development mode. ¯\_(ツ)_/¯

app/models/component_inventory/user.rb
jmervine: Yeah, but (a) I'd need to `with_indifference_access`, which I avoid like the plague (it's super slow and clunky) and (b) I want to be more explicit.

app/models/authentication.rb
jmervine: I try to avoid the bang whenever possible for clarity. When origionally writing this method for Drydock I think I must have renamed it back and forth like 3 or 4 times. I settled on `invalid_email?` because it seemed to fit readability better.

app/admin/user.rb
jmervine: Maybe, this hasn't showed up as an issue in Drydock yet, but it also couldn't hurt.

app/models/component_inventory/base.rb
jmervine: It explodes and reports to rollbar, which is what I want it to do. This workflow is only run as part of a scheduled job. I want the extra visibility.

Also, Rails.cache doesn't cache exceptions:

https://stackoverflow.com/questions/42818442/does-rails-cache-anything-if-an-exception-is-raised

app/controllers/api/v1/chat_ops_controller.rb
```diff
@@ -67,21 +80,22 @@ def slack_event
                         Config.confirmation_reaction_emoji)
 
     render json: t("slack.timeline.added"), status: 200 # Render 200 so slack doesn't retry
-  rescue Slack::Error
-    error = "#{t('slack.errors.timeline')}: #{$ERROR_INFO}"
+  rescue Slack::Error => ex
+    error = "#{t('slack.errors.timeline')}: #{ex}"
```
joshuatobin: /shrug

app/controllers/trello_controller.rb
```diff
@@ -1,41 +1,22 @@
...
-    return redirect_to "/auth/failure" unless current_user
+  def callback
+    auth = Authentication.new(request, provider: :trello)
+    return redirect_to(failure_auth_path) if auth.user.nil?
 
-    update_session_info
-
-    update_current_user_trello_info
-
-    redirect_to session.delete(:return_to)
```
joshuatobin: I think here you're losing the `:return_to`, which in my testing showed that if you tried to create the trello, didn't have auth tokens, it would redirect as expected, update the user model with the tokens, but never redirect back and finish creating the card.

jmervine: Fixed by setting it to `redirect_to` over `return_to` in `app/admin/incident.rb`, see:

https://github.com/heroku/retrodot/pull/270/commits/8b02e305769c3c4446332e32b08e448bd01cf1fc

app/models/authentication.rb
lexelby: Avoid the bang... even to the point of defining a helper method?  To be clear, I meant this:

```
def invalid_email?
  !valid_email?
```

Or if you really REALLY want to avoid a bang:

```
def invalid_email?
  not valid_email?
end
```

My point is to avoid the duplication of that regex.

config/locales/en.yml
lexelby: Use `>` instead, which converts line breaks into spaces and strips indentation ([reference](https://stackoverflow.com/a/3790497/4249120)).

app/models/authentication.rb
jmervine: Ah, I see what you're saying. I didn't realize that `valid_email?` is used, but `invalid_email?` isn't. I'll just remove `invalid_email?` all together. Sorry I thought that only `invalid_email?` was being used / existed. Copy / paste. Need to clean up the drydock `invalid_email?` as well.

config/locales/en.yml
jmervine: thanks for that

db/migrate/20170523225609_rename_google_access_token_column_on_users.rb
jmervine: HA! Wow, just tested this locally. I always assumed that renaming a column dropped the data. It (logically) does not. I updated removing the custom manipulation.

app/models/authentication.rb
jmervine: Couldn't hurt.

rkatti:
👍 looks good

PR #274, opened 2017-06-01
fixing auth edge case

jmervine:
Fixing https://rollbar.com/Heroku-3/retrodot/items/90/


sheax0r:
👍

spec/helpers/active_admin/views_helper_spec.rb
```diff
@@ -1,5 +1,5 @@
 RSpec.describe ActiveAdmin::ViewsHelper do
-  describe :format_duration do
+  describe "format_duration" do
```
jmervine: Something in the `rails-controller-testing` gem doesn't allow symbols as describe arguments.

spec/controllers/application_controller_spec.rb
```diff
@@ -0,0 +1,49 @@
+require "spec_helper"
+
+RSpec.describe ApplicationController, type: :controller do
+  describe "#current_user" do
+    let(:invalid) { create(:user, handle: nil) }
+    let(:valid)   { create(:user) }
+
+    before do
+      # unstub current_user
+      allow(subject).to receive(:current_user).and_call_original
```
sheax0r: Why do we need this? Is there some sort of override of the `current_user` method in spec helpers..?

jmervine: This simple resets `current_user`. It shouldn't be required, but I'm pretty sure (when building the spec I copied this from months ago) I was getting odd errors with stuck stubs.

PR #278, opened 2017-06-02
show correct incident ID for timelines

### LINKED ISSUE #277: Displaying the timeline for the wrong incident ID
Currently, the timeline link for every incident on retrodot's landing page is going to the wrong incident.  As an example, the link for incident 1144 is to incident 1067.

lexelby:
Fixes #277.  It was just cosmetic, and the correct timeline was being displayed.


jmervine:
:+1:

